PR #70, opened 2021-01-12
Improve facet strings distribution aggregation speed

Kerollmops:
I already attempted to speed up the facet strings by developing the facet levels for strings too. Facet levels for numbers are easy to create groups, the left and the right key are encoded to be global orderable. Unfortunately, it is not that easy for strings as they don't have a size known at compile-time we need to pad the left string to make sure that the end of the left string is not compared with the start of the right string or vice-versa.

I closed the pull request https://github.com/Kerollmops/heed/pull/90 I previously made to introduce the custom key comparison function of LMDB as it introduces more problems than I first thought about.

- [ ] Remove [the ugly limit on the number of facet entries to visit](https://github.com/meilisearch/milli/pull/66/commits/83c2150da6bde68436122bc69c0098e7e38635e6).
- [ ] Introduce facet levels for facet strings to speed-up aggregation (we need to restrict the facet string values length to ~256 bytes).

-----

Facet strings are a little bit special compared to facet numbers as the facet value doesn't have a constant and known size, strings length are variable. Facet levels for facet numbers are easy to implement: when we encode a level higher than 0 we simply to the left and right bounds that it defines, one after the other in terms of bytes, when we need to read it, we read two 64bits length slices to extract the low and high bounds. It is not only easy in terms of encoding/decoding but also for LMDB, as we don't have to do anything more for LMDB to sort them in low bound order.

When it comes to strings it is a little bit more complex, we need to:
 - encode the size of, at least, the low bound string,
 - restrict the facet value by half (from ~512 to ~256 bytes) to be able to write two of them (the low and high bounds),
 - use a custom key comparison function with LMDB to help it sort the entries by low bound them high bound (like for numbers).

#### About the custom key comparison function for facet values

If we use a custom key comparison function for the facet string values we will encounter a little problem, facet entries are all in the same LMDB database, therefore the custom key comparison function must be smart enough to understand the facet value type that it is comparing (either a string and i64 or an f64).

There are two solutions to be able to handle that correctly, I listed those by order of preference:
  1. Introduce three facet values documents ids databases for each facet type (strings, i64, and f64).
     We will then be able to set the right custom key comparison function for the right database, it will also reduce [a lot of boilerplate](https://github.com/meilisearch/milli/pull/66/files#diff-ff75c7b3168e9cb19526183b93b2e79ffe7143d1991e8431afb916be696be71bR53-R55) where we were required to remap the key codec based on the facet type.
  2. Append or prepend a byte to the facet values to encode the facet type.
  3. Pad the first string with zeroes to make sure we don't compare the end of the left string with the start of the right string. Doing this has the disadvantage of making the LMDB keys to be at least 250 bytes, the facet string length limit.


ManyTheFish:
PR outdated, we should redesign it.

PR #100, opened 2021-03-04
Improve the Asc/Desc criteria

Kerollmops:
This PR changes:
 - The debug prints to show the name of the ascending/descending field.
 - Reworked the criteria to makes good use of the facet iterators.
    Now instead of going through the whole list of facet values to reach the first candidates according to the facet and order, it keeps the iterator between iterations and pops from it every time until it becomes empty, that's much better 🎉

### DIFF
--- a/milli/src/search/criteria/asc_desc.rs
+++ b/milli/src/search/criteria/asc_desc.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 use std::mem::take;
 
-use anyhow::bail;
+use anyhow::{bail, Context as _};
 use itertools::Itertools;
 use log::debug;
 use ordered_float::OrderedFloat;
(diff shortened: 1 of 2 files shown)


milli/src/search/criteria/asc_desc.rs
```diff
@@ -140,23 +147,28 @@ impl<'t> AscDesc<'t> {
...
         loop {
-            debug!("Facet {} iteration ({:?})",
-                if self.ascending { "Asc" } else { "Desc" }, self.candidates,
+            debug!("Facet {}({}) iteration",
+                if self.ascending { "Asc" } else { "Desc" }, self.field_name
             );
 
-            match &mut self.candidates {
-                candidates if candidates.is_empty() => {
+            match self.candidates.next().transpose()? {
```
LegendreM, line 154: Why transpose?

LegendreM (approved):

PR #101, opened 2021-03-05
implement display for criterion

MarinPostma:
Implement `Display` for `Criterion`.

Needed to return the criterion with the correct formating to the user.

### DIFF
--- a/milli/src/criterion.rs
+++ b/milli/src/criterion.rs
@@ -76,8 +76,8 @@ impl fmt::Display for Criterion {
             Attribute       => f.write_str("attribute"),
             WordsPosition   => f.write_str("wordsPosition"),
             Exactness       => f.write_str("exactness"),
-            Asc(attr)       => write!(f, "asc({:?})", attr),
-            Desc(attr)      => write!(f, "desc({:?})", attr),
+            Asc(attr)       => write!(f, "asc({})", attr),
+            Desc(attr)      => write!(f, "desc({})", attr),
         }
     }
 }
--- a/milli/src/criterion.rs
+++ b/milli/src/criterion.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::fmt;
 
 use anyhow::{Context, bail};
 use regex::Regex;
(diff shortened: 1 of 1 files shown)


milli/src/criterion.rs
```diff
@@ -63,3 +64,20 @@ pub fn default_criteria() -> Vec<Criterion> {
...
+
+        match self {
+            Typo            => write!(f, "typo"),
+            Words           => write!(f, "words"),
+            Proximity       => write!(f, "proximity"),
+            Attribute       => write!(f, "attribute"),
+            WordsPosition   => write!(f, "wordsPosition"),
+            Exactness       => write!(f, "exactness"),
+            Asc(attr)       => write!(f, "asc({:?})", attr),
+            Desc(attr)      => write!(f, "desc({:?})", attr),
```
Kerollmops, line 80:
```suggestion
            Typo            => f.write_str("typo"),
            Words           => f.write_str("words"),
            Proximity       => f.write_str("proximity"),
            Attribute       => f.write_str("attribute"),
            WordsPosition   => f.write_str("wordsPosition"),
            Exactness       => f.write_str("exactness"),
            Asc(attr)       => write!(f, "asc({:?})", attr),
            Desc(attr)      => write!(f, "desc({:?})", attr),
```

Kerollmops (changes requested):

milli/src/criterion.rs
MarinPostma: I don't mind changing, but why? I find it somewhat nicer when they are all the same :)

Kerollmops: Just to skip the formatting part, as it is not necessary :)

milli/src/criterion.rs
```diff
@@ -63,3 +64,20 @@ pub fn default_criteria() -> Vec<Criterion> {
...
+
+        match self {
+            Typo            => f.write_str("typo"),
+            Words           => f.write_str("words"),
+            Proximity       => f.write_str("proximity"),
+            Attribute       => f.write_str("attribute"),
+            WordsPosition   => f.write_str("wordsPosition"),
+            Exactness       => f.write_str("exactness"),
+            Asc(attr)       => write!(f, "asc({:?})", attr),
+            Desc(attr)      => write!(f, "desc({:?})", attr),
```
Kerollmops, line 80:
```suggestion
            Asc(attr)       => write!(f, "asc({})", attr),
            Desc(attr)      => write!(f, "desc({})", attr),
```

Kerollmops (changes requested):

PR #116, opened 2021-03-11
add index metadata

MarinPostma:
add the creation date and update date in the index. It is very cumbersome to deal with it on the client side.

I have made the choice to udpate the `updated_at` time before the udpate is executed, to make sure it is called unconditionally, even in case of short-circuiting paths.

Added an update on `put_primary_key` too.


Kerollmops (changes requested):
Hey thanks, could you rebase please?

PR #118, opened 2021-03-15
fix broken offset

MarinPostma:
Fix broken offset. 'skip' was used instead of `take` on the candidate iterator, so `for_each` would consume all of the iterator, instead of only the `offset` first.

fix https://github.com/meilisearch/transplant/issues/82


Kerollmops (approved):
Ho, nice catch, thank you!

PR #123, opened 2021-03-25
select a specific release of the tokenizer instead of using the latests git commit

irevoire:
This allow us to do breaking change to the tokenizer without breaking milli


Kerollmops (approved):
Thank you :)

PR #126, opened 2021-03-29
Stop words

### LINKED ISSUE #111: Implement stop words
Implement stop words to make transplant iso with the current version of MeiliSearch

irevoire:
closes #111 
I'm not sure I updated everything, can you tell me if I missed something?

Also we should not merge this until we release a new version of the tokenizer (look at the Cargo.toml I pushed)

### DIFF
--- a/milli/src/update/index_documents/mod.rs
+++ b/milli/src/update/index_documents/mod.rs
@@ -436,7 +436,7 @@ impl<'t, 'u, 'i, 'a> IndexDocuments<'t, 'u, 'i, 'a> {
             let readers = rayon::iter::repeatn(documents, num_threads)
                 .enumerate()
                 .map(|(i, documents)| {
-                    let stop_words = fst::Set::default();
+                    let stop_words = self.index.stop_words(self.wtxn)?;
                     let store = Store::new(
                         searchable_fields.clone(),
                         faceted_fields.clone(),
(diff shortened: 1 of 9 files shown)


milli/src/update/index_documents/mod.rs
```diff
@@ -436,7 +436,7 @@ impl<'t, 'u, 'i, 'a> IndexDocuments<'t, 'u, 'i, 'a> {
             let readers = rayon::iter::repeatn(documents, num_threads)
                 .enumerate()
                 .map(|(i, documents)| {
-                    let stop_words = fst::Set::default();
+                    let stop_words = self.index.stop_words(self.wtxn)?;
```
Kerollmops, line 439: Could you please fetch the `stop_words` only once outside of this function, for example on line 406, please?

milli/src/update/settings.rs
```diff
@@ -210,6 +220,35 @@ impl<'a, 't, 'u, 'i> Settings<'a, 't, 'u, 'i> {
...
+
+                // we want to re-create our fst
+                if !identical {
+                    // we need to sort it first. TODO: use btreemap
+                    let mut words: Vec<String> = stop_words.iter().cloned().collect();
+                    words.sort();
+                    let fst = fst::Set::from_iter(&words)?;
+                    self.index.put_stop_words(self.wtxn, &fst)?;
+                }
+                Ok(!identical)
```
Kerollmops, line 242: I am not that's most visible way of doing this, maybe constructing the FST first and then comparing the internal bytes for equality (`fst.as_fst().as_bytes()`).

And to simplify the FST creation why can't we just ask for a `BTreeSet` instead of an `HashSet`?

Kerollmops (changes requested):

irevoire:
Reopening the PR from another branch #128

Issue #49, opened 2020-12-11
Introduce a threshold to the asc/desc sorting criteria


Kerollmops:
The current Asc/Desc criteria choose between two ways of doing the job:
 1. If the number of candidates is lower or equal to 1000, it does a simple sort.
 2. If the number of candidates is bigger than 1000, it uses the facet database.

This number (1000) should better be a percentage of the total number of documents. It could be something like 0.1%. This way, we ensure that the best technique is used even if the database grows. #48 helps in retrieving the number of documents and, therefore, speedups computing a percentage.

shekhirin:
Hi! I tried to resolve this issue in https://github.com/meilisearch/milli/pull/127.

Also I've found another place where such way of choosing an algorithm may be applicable, though it may be a miss since I'm new to the codebase.
https://github.com/meilisearch/milli/blob/38b6e8decd59013ef4e8cae4d7fef8ada80d6252/milli/src/search/facet/facet_distribution.rs#L25
https://github.com/meilisearch/milli/blob/38b6e8decd59013ef4e8cae4d7fef8ada80d6252/milli/src/search/facet/facet_distribution.rs#L210-L214

PR #127, opened 2021-03-29
feat(search, criteria): const candidates threshold

shekhirin:
Resolves https://github.com/meilisearch/milli/issues/49

### DIFF
--- a/milli/src/search/criteria/asc_desc.rs
+++ b/milli/src/search/criteria/asc_desc.rs
@@ -19,9 +19,9 @@ use crate::search::WordDerivationsCache;
 use crate::{FieldsIdsMap, FieldId, Index};
 use super::{Criterion, CriterionResult};
 
-/// If the number of candidates is lower or equal to the specified % of total number of documents,
-/// use simple sort. Otherwise, use facet database.
-const CANDIDATES_THRESHOLD: f64 = 0.1;
+/// Threshold on the number of candidates that will make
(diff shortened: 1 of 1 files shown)


Kerollmops (changes requested):
Thank you very much 🙏

I would like to ask you to change the PR a little bit, could you, please, define a const of 1000, named `CANDIDATES_THRESHOLD` as you've already done?

We have two types of algorithms, a linear one, that fetches documents facet values one by one and does a basic sort, and a set-theory-based algorithm that efficiently iterates over the facet values.

The goal of the threshold is to help us choose between both algorithms, it is much more correct to use the linear algorithm on a fixed number of documents instead of a ratio of the number of documents in the database.

shekhirin:
Ok, got it. So do I get it right that issue https://github.com/meilisearch/milli/issues/49 doesn't make much sense? 😄

Kerollmops:
@shekhirin Yeah, indeed, it is quite outdated. It was created a long time ago and I didn't take the time to update it, sorry for that. Anyway, your PR will close it :)

shekhirin:
@Kerollmops ok, no problem! :) Thanks for clarification. Btw, I've updated the PR with const threshold definition.

Kerollmops (approved):
This is good for me, thank you for your contribution.

Hope to see you on another PR, I will create other "first good issues" issues 😃

Issue #113, opened 2021-03-11
Implement stats


curquiza:
Implement stats (general and of an index) so that transplant can have an iso behavior with the current version of MeiliSearch

More information on [the MeiliSearch /stats reference documentation](https://docs.meilisearch.com/reference/api/stats.html#stats).

Kerollmops:
There are two ways of implementing the `fieldsDistribution` part of the stats route, the simple and the little more complex way.

## The simple way

Don't keep any state and recompute the fields distribution every time a user request the stats. It is a simple as iterating over the index document database, [reading the obkvs](https://docs.rs/obkv/) and increasing a counter for each one of the fields present. This approach can be slow when the number of documents becomes important, but it is a valid implementation.

```rust
impl Index {
    fn stats(&self, &rtxn: &RoTxn) -> anyhow::Result<Stats> {

        for result in self.documents.iter(rtxn)? {
            let (_id, obkv) = result?;
            // fetch and increase a counter for every field in this document (obkv)
        }

        unimplemented!()
    }
}
```

## The little more complex way

This implementation version is as simple as computing the `fieldsDistribution` stats and storing them into the Index itself after every update, an even better solution would be to do so only for any updates that modify the documents.

This solution could be implemented afterward because it needs more carefulness on where and when we should reset the stored stats.

Kerollmops:
Thank you @shekhirin for taking this issue, you are a hell of a contributor 🎉

It could, indeed, be a good idea to introduce a `fields_distribution` method on `Index` that would return an `HashMap` with the number of documents that have the specified fields. I am not sure if it would be better to return `FieldId` as keys or directly the field names (converted by using the `Index::fields_ids_map` method).

For now, we could expose a `field_ids_distribution` method on `Index` that would return an `HashMap` with `FieldId`s as keys and an `u64` for the count, @MarinPostma will convert it itself until he finds it easier than we do it on our side. In this case, we will patch the method. Anyway, that is not exposed to anyone but the MeiliSearch team for now, so we can break things :)

Thank you very much, again!

curquiza:
@Kerollmops can this issue be closed? I mean, does milli return enough information to let transplant implement the `stats/` and `/indexes/:uid/stats` routes?

shekhirin:
Another thing you may want to implement in this repo is calculating the db size. 

I see that previously it was done by traversing the whole db directory
```rust
    let database_size = WalkDir::new(&data.db_path)
        .into_iter()
[...]

PR #135, opened 2021-03-31
feat(index): introduce fields_ids_distribution

shekhirin:
First (and simple) implementation of fields distribution for https://github.com/meilisearch/milli/issues/113

### DIFF
--- a/milli/src/index.rs
+++ b/milli/src/index.rs
@@ -203,6 +203,25 @@ impl Index {
         Ok(self.main.get::<_, Str, SerdeJson<FieldsIdsMap>>(rtxn, FIELDS_IDS_MAP_KEY)?.unwrap_or_default())
     }
 
+    /* fields ids distribution */
+
+    /// Returns the fields ids distribution which associate the internal field ids
+    /// with the number of times it occurs in the obkv documents.
+    // TODO store in the index itself and change only within updates that modify the documents
+    pub fn fields_ids_distribution(&self, rtxn: &RoTxn) -> anyhow::Result<HashMap<FieldId, u64>> {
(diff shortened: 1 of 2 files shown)


milli/src/index.rs
```diff
@@ -203,6 +203,25 @@ impl Index {
         Ok(self.main.get::<_, Str, SerdeJson<FieldsIdsMap>>(rtxn, FIELDS_IDS_MAP_KEY)?.unwrap_or_default())
     }
 
+    /* fields ids distribution */
+
+    /// Returns the fields ids distribution which associate the internal field ids
+    /// with the number of times it occurs in the obkv documents.
+    // TODO store in the index itself and change only within updates that modify the documents
+    pub fn fields_ids_distribution(&self, rtxn: &RoTxn) -> anyhow::Result<HashMap<FieldId, usize>> {
```
Kerollmops, line 211:
```suggestion
    pub fn fields_ids_distribution(&self, rtxn: &RoTxn) -> anyhow::Result<HashMap<FieldId, u64>> {
```

It could be maybe strange but as the documents and document fields are stored on disk, there can be more that 2^32 field occurrences, therefore using an `usize` on a 32bit machine could be a problem, using an u64 is safe.

Kerollmops (changes requested):
Nothing to say, clean and efficient pull request!

Thank you again, you are creating PRs faster than I create issues 😄

milli/src/index.rs
shekhirin: Got it, completely agree. Fixed.

Kerollmops (approved):
Seems good to me, merging :)

PR #136, opened 2021-03-31
feat(index): store fields distribution in index

shekhirin:


shekhirin:
@Kerollmops I just started working on a better implementation of fields ids distribution with caching in index and need your help on one question.

I see there's a bunch of methods such as `update_searchable`, `update_displayed` and `update_facets` which calls the `put_fields_ids_map` method inside. I don't quite get why it's happening. Is it assumed that it's possible to update the searchable/displayed/facets and include previously unknown fields in it? In that case we'll need to call `put_fields_ids_map` to ensure it's up-to-date, yeah, but I don't understand if I get it right.

Kerollmops:
IIRC @MarinPostma and I decided that it was a better idea to store fields name in the database directly, it is easier to work with that when reordering them and fields ids are updated. You can look at how `Index::displayed_fields` work for example and maybe change the `fields_ids_distribution` into a `fileds_distribution` that returns a `HashMap<&str, u64>`.

I hope this answer helps you move forward, tell me otherwise.

shekhirin:
@Kerollmops ok got it. Refactored the distribution to contain names instead of ids.

shekhirin:
@Kerollmops I think I've covered all the places where we need to update the `fields_ids_distribution`. Tell me if I missed anything.

Cargo.lock
```diff
@@ -1520,16 +1520,16 @@ checksum = "d4fd5641d01c8f18a23da7b6fe29298ff4b55afcccdf78973b24cf3175fee32e"
...
 dependencies = [
  "ucd-trie",
 ]
 
 [[package]]
 name = "pest"
 version = "2.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "10f4872ae94d7b90ae48754df22fd42ad52ce740b8f370b03da4835417403e53"
+source = "git+https://github.com/pest-parser/pest.git?rev=51fd1d49f1041f7839975664ef71fe15c7dcaf67#51fd1d49f1041f7839975664ef71fe15c7dcaf67"
```
Kerollmops, line 1532: Why does the lock file change? Have you `cargo update`?

milli/src/index.rs
```diff
@@ -203,23 +206,18 @@ impl Index {
...
-        for document in self.documents.iter(rtxn)? {
-            let (_, obkv) = document?;
+    /* fields distribution */
 
-            for (field_id, _) in obkv.iter() {
-                *distribution.entry(field_id).or_default() += 1;
-            }
-        }
+    /// Writes the fields distribution which associate the field with the number of times
+    /// it occurs in the obkv documents.
```
Kerollmops, line 212:
```suggestion
    /// Writes the fields distribution which associates every field name with
    /// the number of times it occurs in the documents.
```

milli/src/index.rs
```diff
@@ -203,23 +206,18 @@ impl Index {
...
+    pub fn put_fields_distribution(&self, wtxn: &mut RwTxn, distribution: &FieldsDistribution) -> heed::Result<()> {
+        self.main.put::<_, Str, SerdeJson<FieldsDistribution>>(wtxn, FIELDS_DISTRIBUTION_KEY, &distribution)
```
Kerollmops, line 214:
```suggestion
        self.main.put::<_, Str, SerdeJson<FieldsDistribution>>(wtxn, FIELDS_DISTRIBUTION_KEY, distribution)
```

milli/src/index.rs
```diff
@@ -203,23 +206,18 @@ impl Index {
...
+    }
 
-        Ok(distribution)
+    /// Returns the fields distribution which associate the field with the number of times
+    /// it occurs in the obkv documents.
```
Kerollmops, line 218:
```suggestion
    /// Returns the fields distribution which associates every field name with
    /// the number of times it occurs in the documents.
```

milli/src/index.rs
```diff
@@ -452,40 +450,34 @@ impl Index {
...
-        let fields_ids_map = index.fields_ids_map(&rtxn).unwrap();
-
-        let fields_ids_distribution = index.fields_ids_distribution(&rtxn).unwrap();
-        assert_eq!(fields_ids_distribution.len(), 2);
-        assert_eq!(fields_ids_distribution.get(&fields_ids_map.id("age").unwrap()), Some(&1));
-        assert_eq!(fields_ids_distribution.get(&fields_ids_map.id("name").unwrap()), Some(&2));
+        let fields_distribution = index.fields_distribution(&rtxn).unwrap();
+        assert_eq!(fields_distribution, hashmap!{
+            "age".to_string() => 1,
+            "name".to_string() => 2
```
Kerollmops, line 480:
```suggestion
            "name".to_string() => 2,
            "age".to_string() => 1,
```

Kerollmops: Just to keep the same order as the documents.

milli/src/index.rs
```diff
@@ -32,6 +33,8 @@ pub const WORDS_PREFIXES_FST_KEY: &str = "words-prefixes-fst";
 const CREATED_AT_KEY: &str = "created-at";
 const UPDATED_AT_KEY: &str = "updated-at";
 
+pub type FieldsDistribution = HashMap<String, u64>;
```
Kerollmops, line 36: Could you please move this type into the lib.rs file, where we declare them? :)

milli/src/update/clear_documents.rs
```diff
@@ -61,3 +63,55 @@ impl<'t, 'u, 'i> ClearDocuments<'t, 'u, 'i> {
...
+    use super::*;
+
+    #[test]
+    fn clear_documents() {
+        let path = tempfile::tempdir().unwrap();
+        let mut options = EnvOpenOptions::new();
+        options.map_size(10 * 1024 * 1024); // 10 MB
+        let index = Index::new(options, &path).unwrap();
+
+        // First we send 3 documents with an id for only one of them.
```
Kerollmops, line 81:
```suggestion
```

milli/src/update/delete_documents.rs
```diff
@@ -9,6 +9,7 @@ use crate::facet::FacetType;
 use crate::{Index, BEU32, SmallString32, ExternalDocumentsIds};
 use crate::heed_codec::facet::{FieldDocIdFacetStringCodec, FieldDocIdFacetF64Codec, FieldDocIdFacetI64Codec};
 use super::ClearDocuments;
+use std::collections::hash_map::Entry;
```
Kerollmops, line 12: Could you please move this import above the first import (`anyhow`) and keep an empty line, in other words: put it at the top of the import and keep it separated with an empty newline, please?

milli/src/update/delete_documents.rs
```diff
@@ -100,6 +103,17 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
...
+                for (field_id, _) in obkv.iter() {
+                    let field_name = fields_ids_map.name(field_id).unwrap();
+                    if let Entry::Occupied(mut entry) = fields_distribution.entry(field_name.to_string()) {
+                        if entry.get() == &1 {
+                            entry.remove();
+                        } else {
+                            *entry.get_mut() -= 1;
+                        }
+                    }
+                }
```
Kerollmops, line 115: I am sure that we could implement a better way to do this by:
 - avoiding retrieving the field names each time we need to access a field in the `fields_distribution` map.
 - creating a HashMap that as the fields ids as keys and increment the count for each deleted document.
 - doing the fields ids conversion to fields names at the end of the loop.
 - doing the clean up after the loop, after the conversion.

This way we reduce the amount of code we add into the main deletion loop (i.e. `or_default() += 1` instead of a match) and the number of lookups in the `fields_ids_map` to fetch the names.

What do you think?

milli/src/update/index_documents/transform.rs
```diff
@@ -148,6 +152,8 @@ impl Transform<'_, '_> {
             // We prepare the fields ids map with the documents keys.
             for (key, _value) in &document {
                 fields_ids_map.insert(&key).context("field id limit reached")?;
+
+                *fields_distribution.entry(key.to_owned()).or_default() += 1;
```
Kerollmops, line 156: We could do the same way as explained before: introducing a HashMap where the key is the field id and to the merge afterward. This way we reduce the number of string creations in this hot loop.

milli/src/update/index_documents/transform.rs
```diff
@@ -307,6 +315,10 @@ impl Transform<'_, '_> {
                 json_buffer.clear();
                 serde_json::to_writer(&mut json_buffer, &field)?;
                 writer.insert(*field_id, &json_buffer)?;
+
+                let field_name = fields_ids_map.name(*field_id).unwrap();
+
+                *fields_distribution.entry(field_name.to_string()).or_default() += 1;
```
Kerollmops, line 321: Same as the two comments above :)

milli/src/update/settings.rs
```diff
@@ -169,7 +169,7 @@ impl<'a, 't, 'u, 'i> Settings<'a, 't, 'u, 'i> {
         Ok(true)
     }
 
-    /// Udpates the index's searchable attributes. This causes the field map to be recomputed to
+    /// Updates the index's searchable attributes. This causes the field map to be recomputed to
```
Kerollmops, line 172: Ahah 😄 Thank you!

Kerollmops (changes requested):
It is again quite good, apart from the part where there can be a little bit too many intensive operations 😃

Cargo.lock
shekhirin: nope, it's the IDE (CLion + Rust plugin). I suppose it sorts it according to some inner rules. Actually I always do `git checkout -f main Cargo.lock` and now forgot do it 😄 

Fixed

milli/src/update/delete_documents.rs
shekhirin: Good idea, agree. Fixed, check it out.

milli/src/update/delete_documents.rs
```diff
@@ -123,6 +131,21 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
...
+            let count = fields_distribution.entry(field_name.to_string()).or_default();
+
+            if count == &count_diff {
+                fields_distribution.remove(field_name);
+            } else {
+                *count -= count_diff;
+            }
+        }
+
+        self.index.put_fields_distribution(self.wtxn, &fields_distribution)?;
```
Kerollmops, line 147: Could you please add a little comment explaining this algorithm, above it, please?

Kerollmops:
```suggestion
        let mut fields_distribution = self.index.fields_distribution(self.wtxn)?;

        for (field_id, count_diff) in fields_ids_distribution_diff {
            let field_name = fields_ids_map.name(field_id).unwrap();
            if let Entry::Occupied(entry) = fields_distribution.entry(field_name.to_string()) {
                match entry.get().checked_sub(count_diff) {
                    Some(0) | None => entry.remove(),
                    Some(count) => entry.insert(count)
                };
            }
        }

        self.index.put_fields_distribution(self.wtxn, &fields_distribution)?;
```

What do you think about its implementation? It reduces the number of lookups in the database. Could you check that it is valid, please? On my side, it is :)

milli/src/update/index_documents/transform.rs
```diff
@@ -269,6 +284,8 @@ impl Transform<'_, '_> {
         let mut uuid_buffer = [0; uuid::adapter::Hyphenated::LENGTH];
         let mut documents_count = 0;
 
+        let mut fields_ids_distribution = HashMap::<FieldId, u64>::default();
```
Kerollmops, line 287:
```suggestion
        let mut fields_ids_distribution = HashMap::new();
```

milli/src/update/index_documents/transform.rs
```diff
@@ -133,6 +137,8 @@ impl Transform<'_, '_> {
...
```
Kerollmops, line 140:
```suggestion
        let mut fields_ids_distribution = HashMap::new();
```

Kerollmops (changes requested):
That's very good thank you!

I would just like you to change some things and we will be good to merge.
Have you tested these stats on your side, maybe?

milli/src/update/index_documents/transform.rs
shekhirin: It was intended 😄 
```
error[E0282]: type annotations needed for `HashMap<u8, V>`
   --> milli/src/update/index_documents/transform.rs:201:24
    |
140 |         let mut fields_ids_distribution = HashMap::new();
    |             --------------------------- consider giving `fields_ids_distribution` the explicit type `HashMap<u8, V>`, with the type parameters specified
...
201 |         for (field_id, count) in fields_ids_distribution {
    |                        ^^^^^ cannot infer type

```

milli/src/update/index_documents/transform.rs
shekhirin: same here

milli/src/update/delete_documents.rs
shekhirin: LGTM and works. Didn't know about `checked_sub()`, nice way to check overflow safely.

shekhirin:
@Kerollmops 
> Have you tested these stats on your side?

nope, just with unit tests. I think we'll gonna know for sure soon when changes to transplant repo with stats route arrive.

milli/src/update/index_documents/transform.rs
Kerollmops: hum, yeah you can only keep this though :)

```rust
let mut fields_ids_distribution = HashMap::<_, u64>::new();
```

Kerollmops:
Yeah you are right, seems legit 😃

milli/src/update/index_documents/transform.rs
```diff
@@ -147,7 +153,9 @@ impl Transform<'_, '_> {
 
             // We prepare the fields ids map with the documents keys.
             for (key, _value) in &document {
-                fields_ids_map.insert(&key).context("field id limit reached")?;
+                let field_id = fields_ids_map.insert(&key).context("field id limit reached")?;
+
+                *fields_ids_distribution.entry(field_id).or_default() += 1;
```
Kerollmops, line 158:
```suggestion
                *fields_ids_distribution.entry(field_id).or_default() += 1u64;
```

milli/src/update/index_documents/transform.rs
```diff
@@ -307,13 +324,20 @@ impl Transform<'_, '_> {
                 json_buffer.clear();
                 serde_json::to_writer(&mut json_buffer, &field)?;
                 writer.insert(*field_id, &json_buffer)?;
+
+                *fields_ids_distribution.entry(*field_id).or_default() += 1;
```
Kerollmops, line 328:
```suggestion
                *fields_ids_distribution.entry(*field_id).or_default() += 1u64;
```

Kerollmops (changes requested):
Could you just change that, without specifying the type on the `HashMap` like before.

I will merge after those changes I think.

milli/src/update/index_documents/transform.rs
shekhirin: I'd rather suggest something like
```rust
let mut fields_ids_distribution: HashMap<_, u64> = HashMap::new();
```

Type inference still isn't possible in your case because of possible zero-sized `document` variable. I mean `for` loop may not happen as well as increment with `1u64` and then `V` type isn't known.

Kerollmops: No, it doesn't work like that, type resolution is done at compile-time, it doesn't depend on whether the documents `for` loop runs or not. But I am maybe misunderstanding something, here?

shekhirin: Ok I think figured it out. It doesn't work that way
```rust
use std::collections::HashMap;

fn main() {
    let mut hashmap = HashMap::new();

    *hashmap.entry(1).or_default() += 1u64;
}
```

but does work that way
```rust
use std::collections::HashMap;

fn main() {
    let mut hashmap = HashMap::new();

    *hashmap.entry(1).or_insert(0) += 1;
}
```

I think that's the reason in the codebase were mentions of `or_insert(0)` construction and no `or_default()`.

shekhirin: fixed

Kerollmops: Ho ok, yeah that's probably the reason why. Thank you for diving into this inference problem!

milli/src/update/index_documents/transform.rs
```diff
@@ -103,6 +106,7 @@ impl Transform<'_, '_> {
             return Ok(TransformOutput {
                 primary_key,
                 fields_ids_map,
+                fields_distribution: fields_distribution,
```
Kerollmops, line 109:
```suggestion
                fields_distribution,
```

Kerollmops (changes requested):
The last change and I will merge your PR!

Thank you very much for the work you do 🎉 😺

milli/src/update/index_documents/transform.rs
shekhirin: yep, done.

Kerollmops (approved):
Thank you very much, merging!

Cargo.lock
shekhirin: @Kerollmops actually it's `cargo`. IDE calls only two commands in cargo: `cargo build` and `cargo metadata`. Both triggers such changes. In fact, it's the order cargo enforces.

```console
➜ milli (main) ✔ git --no-pager diff

➜ milli (main) ✔ cargo check
    Finished dev [unoptimized + debuginfo] target(s) in 0.25s

➜ milli (main) ✗ git --no-pager diff
diff --git a/Cargo.lock b/Cargo.lock
index 8f296f2..91e7245 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1520,7 +1520,8 @@ checksum = "d4fd5641d01c8f18a23da7b6fe29298ff4b55afcccdf78973b24cf3175fee32e"
 [[package]]
 name = "pest"
 version = "2.1.3"
-source = "git+https://github.com/pest-parser/pest.git?rev=51fd1d49f1041f7839975664ef71fe15c7dcaf67#51fd1d49f1041f7839975664ef71fe15c7dcaf67"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "10f4872ae94d7b90ae48754df22fd42ad52ce740b8f370b03da4835417403e53"
 dependencies = [
  "ucd-trie",
 ]
@@ -1528,8 +1529,7 @@ dependencies = [
 [[package]]
 name = "pest"
 version = "2.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "10f4872ae94d7b90ae48754df22fd42ad52ce740b8f370b03da4835417403e53"
+source = "git+https://github.com/pest-parser/pest.git?rev=51fd1d49f1041f7839975664ef71fe15c7dcaf67#51fd1d49f1041f7839975664ef71fe15c7dcaf67"
 dependencies = [
  "ucd-trie",
 ]
```

Issue #144, opened 2021-04-06
Support synonyms in the engine


Kerollmops:
Milli must support the synonyms. Synonyms can be more than one word and are already supported in the query tree construction logic, we only need to introduce methods to set and save them on disk.

You can find more information on [the official MeiliSearch documentation](https://docs.meilisearch.com/reference/api/synonyms.html#synonyms).

1. Introduce a field in [the Settings struct to accept a list of synonyms](https://github.com/meilisearch/milli/blob/2bcdd8844c4ec9f6f8a34617ea0e4321fa633c0c/milli/src/update/settings.rs#L29-L35).
1. Store this list of synonyms in an efficient way in the key-value store.
1. Fetch this list in [the Context trait implementation of the `QueryTreeBuilder`](https://github.com/meilisearch/milli/blob/2bcdd8844c4ec9f6f8a34617ea0e4321fa633c0c/milli/src/search/query_tree.rs#L188-L190).

## Details about the synonyms

As explained above the synonyms can only be declared one-way (i.e. android gives smartphone), if a user needs to declare a two-way synonym (i.e. wolverine gives logan and logan gives wolverine) it must be declared in both way by the user in the settings, that's expected behavior.

Here is the kind of JSON object that the engine will receive describing the list of synonyms:

```json
{
  "wolverine": ["logan"],
  "logan": ["wolverine"],
  "android": ["smartphone"],
  "wow": ["world of warcraft"],
  "world of worcraft": ["wow"]
}
```

### How can we store the synonyms?

We must find an efficient way to store the synonyms in LMDB (using heed), preferably in the main (polymorph) database like for stop-words. A HashMap serialized using serde JSON could do the job.

However, if a user has a lot of synonyms (think plurals or CJK), it could take a lot of time to deserialize the whole HashMap to, then, only fetch one list of associated synonyms. Finding a better way to store the synonyms (using another database) will be part of another issue, in the future.

### Normalize the synonyms!

When performing a search, the user query is compared to the list of synonyms, for synonyms to be matched correctly with the query words, both synonyms and query words must be normalized. Query words are already normalized (@LegendreM could you confirm?), we must not forget this step with synonyms when storing them. Both sides must be normalized (i.e. wolverine and logan).

If a synonym is found to be multiple words, it must be split using the tokenizer and stored splitter, on both sides.

shekhirin:
> the official MeiliSearch documentation.

I suppose link to the doc isn't correct 😄 
[...]

PR #146, opened 2021-04-07
Facet float-integer becomes facet number

### LINKED ISSUE #142: Only two facet types: string and number
_Related to this specification: https://github.com/meilisearch/specifications/pull/24_

Milli needs to know the facet types.
Currently, Milli handles 3 types: float, integer and string.

Milli will handle 2 types only: number and string.
Numbers are any integers or floats, and these will be converted into float.

Kerollmops:
This pull request fixes #142 by converting all the integers into floats, introducing a new "number" type. The exact precision for integers converted into [64bits floats is between -2^53 and 2^53](https://en.wikipedia.org/wiki/Double-precision_floating-point_format#Precision_limitations_on_integer_values), it has been decided that it was enough for common use cases and ease the search engine usage.

More information can be found [in this specification draft document](https://hackmd.io/@WZKoavtcShyXvCB_qthGRQ/Hka6af-Vu).

- [x] Convert booleans into the appropriate facet type when declared. Keep in mind that the original document is never modified, the document is always returned as it was received.
- [x] Don't skip empty strings and consider them as valid facet values.
- [x] Remove trait bounds from the facet functions and types (`FacetIter`, `FacetRange`) to work only for numbers (`f64`).
- [x] Warn instead of raising an error and stop the indexing when a conversion to the facet type is impossible.

---

In a future pull request, we should split both facet databases of strings and floats. Doing so would make it possible to use a different implementation of facet levels for strings and numbers, strings can't work like numbers due to their inherent unknown size making facet distribution aggregation for strings slow. Then wrap both facet databases into a type that would help us work with facets strings and numbers. There can be helping methods to return iterators on ranges of numbers, iterators for facet ranges...


curquiza:
Will this PR fix completely https://github.com/meilisearch/milli/issues/142 or this issue will need another PR?

Kerollmops:
@curquiza Yep, this fixes #144.

Kerollmops:
Hey @MarinPostma, Could you please check that the distinct system is still valid? Thank you :)

milli/src/heed_codec/facet/mod.rs
```diff
@@ -1,13 +1,9 @@
 mod facet_level_value_f64_codec;
-mod facet_level_value_i64_codec;
 mod facet_value_string_codec;
 mod field_doc_id_facet_f64_codec;
-mod field_doc_id_facet_i64_codec;
 mod field_doc_id_facet_string_codec;
 
 pub use self::facet_level_value_f64_codec::FacetLevelValueF64Codec;
```
MarinPostma, line 6: Maybe we should change this to `FacetLevelValueNumberCodec`

MarinPostma (changes requested):
Looks good to me on the distinct side.

So no inference for now, we just drop the `Integer` and `Float` facet types and replace them with `Number`?

milli/src/heed_codec/facet/mod.rs
Kerollmops: As this is a low-level codec implementation that encodes and decodes f64 directly, I prefer keeping the explicit name.

Kerollmops:
> So no inference for now, we just drop the `Integer` and `Float` facet types and replace them with `Number`?

The inference is no more needed, it has been decided that there are only soft type conversions, strings will be moved into the string database and JSON numbers will be moved in the f64 database.

PR #172, opened 2021-04-28
Optimize proximity criterion

LegendreM:
Add a threshold on proximity when choosing between linear and set algorithm


Kerollmops (approved):
Perfect thank you, this will speed up the criteria a lot!

Issue #112, opened 2021-03-11
Implement distinctAttribute


curquiza:
Implement `distinctAttributes`.

The application will be after the bucket sort for this first iteration.

Giving the possibility to apply the distinct before the bucket sort might be implemented once the new search engine is out and needs a new issue.

curquiza:
To be iso, the users should be able to distinct without pass the field in the facets.
However, if they want performance, the users should pass the facets in the faceted attributes (in the settings).

PR #173, opened 2021-04-28
Remove excluded document in criteria iterations

### LINKED ISSUE #170: Documents are sometimes returned multiple times
We found a bug in the engine that was sitting here for some time now, we think that it was introduced in #106.

We have found a solution to this issue, and it is a lot what we wanted to implement for the optimized version of the distinct system. We want to aggregate the whole list of candidates that we already returned from the criteria loop, combine them with the forbidden candidates that have been removed by distinct and, give this list to the criteria to make sure that we don't return candidates more than once and also be sure that we don't lose performances by going up and down in the criteria list but stopping early.

I think it could be a great idea to let @MarinPostma and @LegendreM work on improving the distinct system by implementing this algorithm and plug the patch on it.

LegendreM:
- pass excluded document to criteria to remove them in higher levels of the bucket-sort
- merge already returned document with excluded documents to avoid duplicas

Related to #125 and #112
Fix #170

### DIFF
--- a/milli/src/search/criteria/final.rs
+++ b/milli/src/search/criteria/final.rs
@@ -5,7 +5,7 @@ use roaring::RoaringBitmap;
 
 use crate::search::query_tree::Operation;
 use crate::search::WordDerivationsCache;
-use super::{resolve_query_tree, Criterion, CriterionResult, Context};
+use super::{resolve_query_tree, Criterion, CriterionResult, CriterionParameters, Context};
 
 /// The result of a call to the fetcher.
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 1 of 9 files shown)


Kerollmops (changes requested):
Thank you, I will merge that tomorrow!

MarinPostma:
Maybe I'm missing something, but aren't candidates passed from a criterion to the next? If it's the case, isn't it necessary to  exclude the documents only on the first criterion in the chain?

LegendreM:
@MarinPostma we could do as you say. But doing it at each call of next avoid each criterion to fetch and return already distinct documents.

milli/src/search/criteria/final.rs
```diff
@@ -22,27 +22,39 @@ pub struct Final<'t> {
...
-                        (Some(qt), None) => resolve_query_tree(self.ctx, qt, &mut HashMap::new(), &mut self.wdcache)?,
-                        (None, None) => self.ctx.documents_ids()?,
+                    let candidates = match candidates {
+                        Some(candidates) => candidates,
+                        None => {
+                            let candidates = match query_tree.as_ref() {
+                                Some(qt) => resolve_query_tree(self.ctx, qt, &mut HashMap::new(), &mut self.wdcache)?,
+                                None => self.ctx.documents_ids()?,
+                            };
+                            bucket_candidates.union_with(&candidates);
```
Kerollmops, line 52:
```suggestion
                            bucket_candidates |= &candidates;
```

milli/src/search/criteria/final.rs
```diff
@@ -22,27 +22,39 @@ pub struct Final<'t> {
...
+                            candidates
+                        }
                     };
 
-                    bucket_candidates.union_with(&candidates);
+                    self.returned_candidates.union_with(&candidates);
```
Kerollmops, line 57:
```suggestion
                    self.returned_candidates |= &candidates;
```

milli/src/search/criteria/typo.rs
```diff
@@ -134,7 +148,7 @@ impl<'t> Criterion for Typo<'t> {
                         Some(CriterionResult { query_tree, candidates, bucket_candidates }) => {
                             self.query_tree = query_tree.map(|op| (maximum_typo(&op), op));
                             self.number_typos = 0;
-                            self.candidates = candidates.map_or_else(Candidates::default, Candidates::Allowed);
+                            self.candidates = candidates.map_or_else(|| Candidates::Forbidden(params.excluded_candidates.clone()), Candidates::Allowed);
```
Kerollmops, line 151:
```suggestion
                            self.candidates = candidates.map_or_else(|| {
                                Candidates::Forbidden(params.excluded_candidates.clone())
                            }, Candidates::Allowed);
```

Kerollmops (changes requested):

Kerollmops (approved):
Thank you, merging!

Issue #175, opened 2021-04-29
CSV delimiter not recognized


yanns:
I'm trying the example given in the README:
```
$ echo "name,age\nhello,32\nkiki,24\n" | http POST 127.0.0.1:9700/documents content-type:text/csv
```

But this does not create any document.

https://github.com/meilisearch/milli/blob/2658c5c545489dc1f0e177a9e3128153aa6bc6a7/milli/src/update/index_documents/transform.rs#L233

returns as headers `StringRecord(["name", "age\\nhello", "32\\nkiki", "24\\n"])`. I suppose it should return `StringRecord(["name", "age"])` instead.

The `\n` seems to be changed in `\\n` (or it is only the effect of debugging)

To be noted that I'm testing on mac.

yanns:
To test, if I change the code to:
```
        let content = "name,age\nhello,32\nkiki,24\n";
        let mut csv = csv::Reader::from_reader(content.as_bytes());
```

then it's working as expected.

yanns:
Arg, the issue is coming from `echo`
Using `printf` fixes the issue:
```
printf "name,age\nhello,32\nkiki,24\n" | http POST 127.0.0.1:9700/documents content-type:text/csv --
```

PR #176, opened 2021-04-29
do not use echo that espaces newline

yanns:
Fix https://github.com/meilisearch/milli/issues/175


Kerollmops (approved):
Thank you very much for your PR, I didn't check that line, my bad!

PR #178, opened 2021-04-29
make document addition number visible

MarinPostma:
The `nb_document` needs to be public to allow the conversion of update statuses to the legacy meilisearch format.


Kerollmops (approved):
Whoops, it seems like something I forgot to made public.

PR #185, opened 2021-05-04
Provide an iterator over all the documents in a milli index

irevoire:


milli/src/index.rs
```diff
@@ -477,6 +477,18 @@ impl Index {
...
+    /// Returns an iterator over all the documents in the index.
+    pub fn all_documents<'t>(
+        &self,
+        rtxn: &'t RoTxn,
+    ) -> anyhow::Result<impl Iterator<Item = heed::Result<(DocumentId, obkv::KvReader<'t>)>>> {
+        Ok(self
+            .documents
+            .iter(rtxn)?
+            // we cast the BEU32 to a DocumentId
+            .map(|document| document.map(|(id, obkv)| (id.get(), obkv))))
```
Kerollmops, line 489:
```suggestion
        // we cast the BEU32 to a DocumentId
        Ok(self.documents.iter(rtxn)?.map(|result| result.map(|(id, obkv)| (id.get(), obkv))))
```

Kerollmops (changes requested):
Yeah, that's great, thank you!

milli/src/index.rs
irevoire: are you sure you want to do that? RustFmt is not happy with this change 🥺

Kerollmops (approved):
Thank you, RustFmt is better than me at formatting :)

bors merge

PR #193, opened 2021-05-06
Fix primary key behavior

### LINKED ISSUE #194: Wrong primary key returned when doing inference
Transplant does not return the correct primary key when doing inference during document addition.

Example:

Add this documents
```json
[
  { "objectId": 2, "title": "Pride and Prejudice" }
]
```

Should give me this result when I get the index info:

```json
{
    "uid": "indexUID",
    "name": "indexUID",
    "createdAt": "2021-04-27T15:41:13.439646200Z",
    "updatedAt": "2021-04-27T15:41:23.465335700Z",
    "primaryKey": "objectId"
}
```

But I get:
```
{
    "uid": "indexUID",
    "name": "indexUID",
    "createdAt": "2021-04-27T15:41:13.439646200Z",
    "updatedAt": "2021-04-27T15:41:23.465335700Z",
    "primaryKey": "id"
}
```

Version: alpha4

MarinPostma:
this pr:
- Adds early returns on empty document additions, avoiding error messages to be returned when adding no documents and no primary key was set.
- Changes the primary key inference logic to match that of legacy meilisearch.

close #194

### DIFF
--- a/milli/src/update/index_documents/mod.rs
+++ b/milli/src/update/index_documents/mod.rs
@@ -1,7 +1,7 @@
 use std::borrow::Cow;
 use std::collections::HashSet;
 use std::fs::File;
-use std::io::{self, Seek, SeekFrom};
+use std::io::{self, Seek, SeekFrom, BufReader, BufRead};
 use std::num::{NonZeroU32, NonZeroUsize};
 use std::str;
 use std::sync::mpsc::sync_channel;
@@ -326,6 +326,16 @@ impl<'t, 'u, 'i, 'a> IndexDocuments<'t, 'u, 'i, 'a> {
         R: io::Read,
         F: Fn(UpdateIndexingStep, u64) + Sync,
     {
+        let mut reader = BufReader::new(reader);
+        reader.fill_buf()?;
+
+        // Early return when there are no document to add
+        if reader.buffer().is_empty() {
+            return Ok(DocumentAdditionResult {
+                nb_documents: 0,
+            })
+        }
+
         self.index.set_updated_at(self.wtxn, &Utc::now())?;
         let before_transform = Instant::now();
         let update_id = self.update_id;
--- a/milli/src/update/index_documents/mod.rs
+++ b/milli/src/update/index_documents/mod.rs
@@ -329,7 +329,7 @@ impl<'t, 'u, 'i, 'a> IndexDocuments<'t, 'u, 'i, 'a> {
         let mut reader = BufReader::new(reader);
         reader.fill_buf()?;
 
-        // Early return when there are no document to add
+        // Early return when there is no document to add
         if reader.buffer().is_empty() {
             return Ok(DocumentAdditionResult {
                 nb_documents: 0,
--- a/milli/src/update/index_documents/transform.rs
+++ b/milli/src/update/index_documents/transform.rs
@@ -95,7 +95,11 @@ impl Transform<'_, '_> {
 
         // We extract the primary key from the first document in
         // the batch if it hasn't already been defined in the index
-        let first = documents.peek().and_then(|r| r.as_ref().ok());
+        let first = match documents.peek().map(Result::as_ref).transpose() {
+            Ok(first) => first,
+            Err(_) => return Err(documents.next().unwrap().unwrap_err().into()),
+        };
+
         let alternative_name = first.and_then(|doc| doc.keys().find(|f| is_primary_key(f)).cloned());
         let (primary_key_id, primary_key) = compute_primary_key_pair(
             self.index.primary_key(self.rtxn)?,
@@ -236,7 +240,7 @@ impl Transform<'_, '_> {
                // The primary key is known so we must find the position in the CSV headers.
                headers.iter().position(|h| h == primary_key)
             },
-            None => headers.iter().position(|f| is_primary_key(&f)),
+            None => headers.iter().position(is_primary_key),
         };
 
         // Returns the field id in the fields ids map, create an "id" field
--- a/milli/src/update/index_documents/transform.rs
+++ b/milli/src/update/index_documents/transform.rs
@@ -47,6 +47,10 @@ pub struct Transform<'t, 'i> {
     pub autogenerate_docids: bool,
 }
 
+fn is_primary_key(field: impl AsRef<str>) -> bool {
+    field.as_ref().to_lowercase().contains(DEFAULT_PRIMARY_KEY_NAME)
+}
+
 impl Transform<'_, '_> {
     pub fn output_from_json<R, F>(self, reader: R, progress_callback: F) -> anyhow::Result<TransformOutput>
     where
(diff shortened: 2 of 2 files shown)


MarinPostma:
closes #194

milli/src/update/index_documents/mod.rs
```diff
@@ -326,6 +326,16 @@ impl<'t, 'u, 'i, 'a> IndexDocuments<'t, 'u, 'i, 'a> {
...
+        let mut reader = BufReader::new(reader);
+        reader.fill_buf()?;
+
+        // Early return when there is no document to add
+        if reader.buffer().is_empty() {
+            return Ok(DocumentAdditionResult {
+                nb_documents: 0,
+            })
+        }
+
```
Kerollmops, line 338: Is it a way to make sure that we do not create a default primary key that is "id" because no documents (and therefore no primary key) were found?

MarinPostma: No, it's a way to not throw a "could not infer primary key" on an empty addition 🙂

MarinPostma: Or is it? 🤔 I'm tired, maybe I understood @irevoire issue wrong

irevoire: Yeah that was my issue

MarinPostma: Tell me if this solves it ☺️ (next week)

milli/src/update/index_documents/transform.rs
```diff
@@ -92,7 +96,7 @@ impl Transform<'_, '_> {
         // We extract the primary key from the first document in
         // the batch if it hasn't already been defined in the index
         let first = documents.peek().and_then(|r| r.as_ref().ok());
```
Kerollmops, line 98: I found this line to make the engine return an error that is not the right one, when there is a decoding error we simply ignore it (`.ok()`), and then this a "missing primary key" error that is thrown due to the fact that no primary key was guessed **and** that the `alternate_name` is None because the first document is invalid.

We should early return when we find that the first document is erroneous, it would be better, but the error doesn't impl `Clone` so we can't return it from the `peek`. We should check that the peeked document is `Err` and if it is we should next and return the error (`unwrap_err`).

milli/src/update/index_documents/transform.rs
```diff
@@ -92,7 +96,7 @@ impl Transform<'_, '_> {
...
-        let alternative_name = first.and_then(|doc| doc.keys().find(|k| k.contains(DEFAULT_PRIMARY_KEY_NAME)).cloned());
+        let alternative_name = first.and_then(|doc| doc.keys().find(|f| is_primary_key(f)).cloned());
```
Kerollmops, line 99:
```suggestion
        let alternative_name = first.and_then(|doc| doc.keys().find(is_primary_key).cloned());
```

milli/src/update/index_documents/transform.rs
```diff
@@ -232,7 +236,7 @@ impl Transform<'_, '_> {
                // The primary key is known so we must find the position in the CSV headers.
                headers.iter().position(|h| h == primary_key)
             },
-            None => headers.iter().position(|h| h.contains("id")),
+            None => headers.iter().position(|f| is_primary_key(&f)),
```
Kerollmops, line 239:
```suggestion
            None => headers.iter().position(is_primary_key),
```

Kerollmops (changes requested):

milli/src/update/index_documents/transform.rs
LegendreM:
```suggestion
        let first = match iter.peek() {
            Some(Err(_)) => iter.next().transpose()?,
            Some(Ok(x)) => Some(x),
            None => None,
        };
```

Untested in this context

ManyTheFish (dismissed):

milli/src/update/index_documents/mod.rs
MarinPostma: I have a client side fix for this, but I am not sure whose duty it is to do that, milli's or the clien?

Kerollmops: But doesn't this commit fix the issue, here?

MarinPostma: yes but I ended up implementing a fix in dumps so we could merge it. We can leave it here, and 'll remove it from dumps. The question is, should you get an error for pushing an empty document addition or not? I claim that we shouldn't, but I wanted to know if you thought otherwise.

milli/src/update/index_documents/transform.rs
MarinPostma: not possible, f is a &&String, and rustc doesn't like it.

milli/src/update/index_documents/transform.rs
MarinPostma: this doesn't compile, but I wrote something similar.

milli/src/update/index_documents/mod.rs
Kerollmops: You must not get an error if you push an empty update, that's valid, even for the first update (where we guess the primary key).

milli/src/update/index_documents/transform.rs
```diff
@@ -91,8 +95,12 @@ impl Transform<'_, '_> {
 
         // We extract the primary key from the first document in
         // the batch if it hasn't already been defined in the index
-        let first = documents.peek().and_then(|r| r.as_ref().ok());
-        let alternative_name = first.and_then(|doc| doc.keys().find(|k| k.contains(DEFAULT_PRIMARY_KEY_NAME)).cloned());
+        let first = match documents.peek().map(Result::as_ref).transpose() {
+            Ok(first) => first,
+            Err(_) => return Err(documents.next().unwrap().unwrap_err().into()),
+
```
Kerollmops, line 101: As we know this is an error we can reduce a little bit the `unwrap`s + `return`s boilerplate.

```suggestion
            Err(_) => documents.next().unwrap()?,
```

Kerollmops (changes requested):
That's fine, thank you for your PR, I merge it after these little changes are applied :)
Thank you!

milli/src/update/index_documents/transform.rs
MarinPostma: that is not gonna work, the first branch is a ref, and next returns a owned

Kerollmops: Ho yeah, you are right, my badu!

Kerollmops: But please remove this newline :)

Kerollmops (approved):
Thank you very much, merging 😃
bors merge

Kerollmops:
bors merge

Kerollmops:
bors merge

Kerollmops:
bors merge

PR #202, opened 2021-05-27
Add field id word count docids database

### LINKED ISSUE #165: Implemention for Exactness criterion
# Prior Art 
Algolia explains the Exact criterion like:
> Records with words that match query terms exactly are ranked higher. The more matching words in a record’s attribute, the higher a record is ranked. By default, an exact match occurs when a full word in a query is matched without typos to a word in an attribute. An inexact match is one that has typos or only matches on a prefix.

> Additionally, synonym matching and plural/singular matching are considered exact. Thus, a word is considered an exact match if its synonym matches a query exactly.

Adding some exceptions:
> [...] single-word matches on multi-word attributes aren’t considered exact matches by default [...]
> [...] Multiword synonyms are not counted as exact matches by default [...]

## Brief
for each word of the query we match documents:
- without typo
- without prefix
- without ngrams
- without complex synonyms
- without word split

if the query only contains 1 word, only 1-word attributes can match.

## Pro & Con

This behavior of the Exact criterion gives a better rank to documents that match the words of the original query, ranking `ngrams`, `complex synonyms`, and `word split` after.
But the exception `if the query only contains 1 word, only 1-word attributes can match` will only boost documents that have an attribute that exactly contains the query word and that's all. No boost on non `complex synonyms`, and non `word split`.
Useful when an ID is searched but in others cases, it could impact the relevancy.
Moreover, when we a multi-words query, attributes that exactly contains the query are not boosted.

At last, because meilisearch is a `search as you type` engine, attributes that exactly begin by the original query could be boosted.

# Possible Behavior in milli

we could divide Exact into several layers of ranking:

## 1: Try to match exactly the query in an attribute
Any document that has an attribute that contains exactly the query:
- without typo
- without prefix
- without ngrams
- without complex synonyms
- without word split
- with words in the right order

has a rank of `0` (perfect match)

## 2: Try to match exactly the query at the start of an attribute
Any document that has an attribute that contains the query at the `n` firsts positions:
- without typo
- without prefix
- without ngrams
- without complex synonyms
- without word split
- with words in the right order

**OR**

Any document that has an attribute that contains exactly the query:
- without typo
- without ngrams
- without complex synonyms
- without word split
- with words in the right order
- **with the last word considered as prefix**

has a rank of `1` (the user didn't finish to tip the search but the final query could be an exact match)

## 3: Try to match exactly the word of the query anywhere in the document
Any other document has a rank of `2` + `1` for each word that is not an exact match.

a word is considered as an exact match when it matches:
- without typo
- without prefix
- without ngrams
- without complex synonyms
- without word split

_**note** @LegendreM: forbidding typos in the `Exact` criterion overlap the work of the `Typo` criterion, we should think about avoiding this conflict by allowing typos in the `Exact` criterion. It seems strange, but, thinking about the interaction between criteria, it could be logical._

### LINKED ISSUE #196: Implement rule 1 for exactness criterion
The exactness criterion is already implemented but the first rule is missing.
See https://github.com/meilisearch/milli/issues/165

Replaces https://github.com/meilisearch/milli/issues/115

LegendreM:
This PR introduces a new database, `field_id_word_count_docids`, that maps the number of words in an attribute with a list of document ids. This relation is limited to attributes that contain less than 11 words.
This database is used by the exactness criterion to know if a document has an attribute that contains exactly the query without any additional word.

Fix #165 
Fix #196
Related to [specifications:#36](https://github.com/meilisearch/specifications/pull/36)

### DIFF
--- a/infos/src/main.rs
+++ b/infos/src/main.rs
@@ -157,6 +157,8 @@ enum Command {
         prefixes: Vec<String>,
     },
 
+    /// Outputs a CSV with the documents ids along with
+    /// the field id and the word count where it appears.
     FieldIdWordCountDocids {
         /// Display the whole documents ids in details.
         #[structopt(long)]
@@ -714,8 +716,8 @@ fn field_id_word_count_docids(
         .id(&field_name)
         .with_context(|| format!("unknown field name: {}", &field_name))?;
 
-    let left = (field_id, 1);
-    let right = (field_id, 11);
+    let left = (field_id, 0);
+    let right = (field_id, u8::max_value());
     let iter = index.field_id_word_count_docids
         .range(rtxn, &(left..=right))?;
 
--- a/infos/src/main.rs
+++ b/infos/src/main.rs
@@ -23,6 +23,7 @@ const WORD_PAIR_PROXIMITY_DOCIDS_DB_NAME: &str = "word-pair-proximity-docids";
 const WORD_PREFIX_PAIR_PROXIMITY_DOCIDS_DB_NAME: &str = "word-prefix-pair-proximity-docids";
 const WORD_LEVEL_POSITION_DOCIDS_DB_NAME: &str = "word-level-position-docids";
 const WORD_PREFIX_LEVEL_POSITION_DOCIDS_DB_NAME: &str = "word-prefix-level-position-docids";
+const FIELD_ID_WORD_COUNT_DOCIDS_DB_NAME: &str = "field-id-word-count-docids";
 const FACET_ID_F64_DOCIDS_DB_NAME: &str = "facet-id-f64-docids";
 const FACET_ID_STRING_DOCIDS_DB_NAME: &str = "facet-id-string-docids";
 const FIELD_ID_DOCID_FACET_F64S_DB_NAME: &str = "field-id-docid-facet-f64s";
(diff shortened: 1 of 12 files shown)


Kerollmops (changes requested):
Could you add this database to the infos crate, this way we can have information on what this database contains.

Kerollmops:
bors try

Kerollmops:
bors try

infos/src/main.rs
```diff
@@ -155,6 +157,15 @@ enum Command {
         prefixes: Vec<String>,
     },
 
+    FieldIdWordCountDocids {
```
Kerollmops, line 160: Could you please add a short description of what this subcommand do?

milli/src/update/delete_documents.rs
```diff
@@ -316,6 +317,20 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
 
         drop(iter);
 
+        // Remove the documents ids from field id word count database.
```
Kerollmops, line 320:
```suggestion
        // Remove the documents ids from the field id word count database.
```

milli/src/search/criteria/exactness.rs
```diff
@@ -162,11 +162,11 @@ fn resolve_state(
     use State::*;
     match state {
         ExactAttribute(mut allowed_candidates) => {
-            let query_len = query.len() as u32;
+            let query_len = query.len() as u8;
```
Kerollmops, line 165: I am not quite fine with this cast are you sure that we will not have more than 256 words in the query?

milli/src/update/index_documents/store.rs
```diff
@@ -58,6 +59,7 @@ pub struct Store<'s, A> {
     // Caches
     word_docids: LinkedHashMap<SmallVec32<u8>, RoaringBitmap>,
     word_docids_limit: usize,
+    field_id_word_count_docids: HashMap<(u8, u8), RoaringBitmap>,
```
Kerollmops, line 62:
```suggestion
    field_id_word_count_docids: HashMap<(FieldId, u8), RoaringBitmap>,
```

Kerollmops (changes requested):

Kerollmops:
BTW, @LegendreM doesn't it fixes #196?

Kerollmops (approved):
bors merge

PR #204, opened 2021-06-01
Decorrelate Distinct, Asc/Desc, Filterable fields from the faceted fields

### LINKED ISSUE #152: Facets and filters behavior
Implement behavior in this spec: https://github.com/meilisearch/specifications/pull/27/files

- [ ] The `facetFilters` parameter during the search is removed. Only `filter` can be used.
- [ ] The users need to set the attributes to `attributesForFaceting` to use the filters during the search via the `filter` parameter. But they don't have to specify the type. Milli must accept only `attributesForFaceting` in an array like this:

```json
{
  "attributesForFaceting": ["author", "price"]
}
```
=> this is the current MeiliSearch v0.20.0 behavior: https://docs.meilisearch.com/reference/api/attributes_for_faceting.html#attributes-for-faceting
=> this means milli has to create two data structures: one for the strings and one for the numbers. No conversions are done in the user documents: `"price": "12"` is not considered as a numeric field and should be in the string data structure.
- [ ] when applying `filter` with a numeric operator, milli should search into the number data structure only. Otherwise in both data structures. More detail in the spec.
- [ ] The `filter` parameter can accept both syntaxes: string (with `OR`/`AND`/`NOT`) and array. An element of an array syntax can accept the string syntax. The first depth of the array syntax is combined with `AND`s, the second depth is combined with `OR`s.

Feel free to complete my issue @Kerollmops with all the technical parts we need to implement these and if some parts need to be implemented on the transplant side 🙂

### LINKED ISSUE #161: asc/desc behavior
_Related to this spec: https://github.com/meilisearch/specifications/pull/30_

According to the spec, here is the behavior of asc/desc:

- [ ] The users will not pass the attribute related to the custom ranking rule into `attributeForFaceting` when setting a ranking rule. It means the search engine must create a database for the related attribute.

ex:
```json
{
  "rankingRules": ["words", "typo", "asc(price)", "proximity"],
  "attributesForFaceting": ["genre"]
}
```
means the search engine has to create facet databases for `genre` and `price`.

- [ ] If the users send documents that don't contain the attributes in the custom ranking rules, the search engine should behave as MeiliSearch v0.20.0: it returns documents that don't contain the attribute at the end of the bucket. This behavior is considered better than discarding the documents. A warning is already added to the docs concerning this specific behavior we call "undefined": https://docs.meilisearch.com/reference/features/settings.html#custom-ranking-rule

- [ ] The search engine is technically able to apply filters and facet distribution on an attribute associated with a ranking rule (thanks to the database creation). However, we should prevent this behavior. To avoid confusion, the search engine should prevent the users to execute a filter or get facet distribution on the ranking attributes. Only the ranking rule should be available for this field (except if the attribute is already present in `attributesForFaceting`).

### LINKED ISSUE #168: Distinct attribute behavior
_Related to this spec: https://github.com/meilisearch/specifications/pull/32/files_

The behavior of the distinct attribute should be identical to the one in MeiliSearch v0.20.0. It means the users don't need to pass the distinct attribute in `attributesForFaceting` even for better performance. The search engine needs to create the appropriate database for this attribute when getting the `distinctAttribute` in the settings.

⚠️ The users should not be able to apply faceting or filtering on this attribute except if they pass it in `attributesForFaceting` as well.

⚠️ This issue should be done once all the criteria are merged.

Kerollmops:
This PR decorrelates the fields that need to be stored in facet databases (big inverted indexes for fast access) from the filterable fields, the previously named faceted fields are now named filterable fields and are the union of the distinct attribute, all the Asc/Desc criteria and, the filterable fields.

I added two tests to make sure that the engine was correctly generating the faceted databases when a distinct attribute or an Asc/Desc criteria were added, and one to make sure that it was impossible to filter on a non-filterable field even if it was a faceted one.

Note that the `AttributesForFacetting` has also been renamed into `FilterableAttributes`. But it will be the Transplant's job to do that on the API, this change is only visible to the milli's library users.

- Related to https://github.com/meilisearch/transplant/issues/187.
- Fixes #161 by returning the documents that don't have the Asc/Desc field at the end of the bucket.
- Fixes #168.
- Fixes #152.


ManyTheFish (approved):

ManyTheFish (approved):

Kerollmops:
bors merge

PR #203, opened 2021-05-31
Make the MatchingWords return the number of matching bytes

LegendreM:
Make the MatchingWords return the number of matching bytes using a custom Levenshtein algorithm.

Fix #138

PR #206, opened 2021-06-01
Fix http-ui

irevoire:
I just noticed that `http-ui` was not compiling on `main`.
I'm not sure this is the best fix, but it works 👀


Kerollmops (approved):
Thank you very, it was indeed introduced by #203.
bors merge

PR #207, opened 2021-06-02
Benchmarks

irevoire:


curquiza (approved):
Good job tamo! We have nice benchmarks now 👏

irevoire:
I'll let @Kerollmops merge, so if it fails, it'll be entirely his responsibility :eyes:

benchmarks/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "benchmarks"
+version = "0.1.0"
+edition = "2018"
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
Kerollmops, line 7:
```suggestion
```

benchmarks/benches/songs.rs
```diff
@@ -0,0 +1,211 @@
...
+        .chain(default_criterion.clone())
+        .collect();
+
+    let basic_with_quote: Vec<String> = BASE_CONF
+        .queries
+        .iter()
+        .map(|s| {
+            s.trim()
+                .split(' ')
+                .map(|s| format!(r#""{}""#, s))
```
Kerollmops, line 74: Doesn't this just do the job?

```suggestion
                .map(|s| format!("{:?}", s))
```

benchmarks/benches/utils.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub configure: fn(&mut Settings),
+    pub facet_condition: Option<&'a str>,
+    /// enable or disable the optional words on the query
+    pub optional_words: bool,
+    /// primary key, if there is None we'll auto-generate docids for every documents
+    pub primary_key: Option<&'a str>,
+}
+
+impl Conf<'_> {
+    fn nop(_builder: &mut Settings) {}
```
Kerollmops, line 32: What's that `nop` method for?

Kerollmops (changes requested):
That' a really good PR, thank you very much, will merge when the changes are applied :)

benchmarks/benches/songs.rs
irevoire: This works, but I don't think we should rely on the result of the `Debug` trait. Right now, it works (and it'll probably work forever), but since they wrote that in the std, I tend to never use it like that:
>>> Derived Debug formats are not stable, and so may change with future Rust versions. Additionally, Debug implementations of types provided by the standard library (libstd, libcore, liballoc, etc.) are not stable, and may also change with future Rust versions.

But we can use it, do as you want :grin:

benchmarks/benches/utils.rs
irevoire: I use it in my `BASE` configuration below.
I could've used an `Option` in the structure, but a no-operation also seems good to me and makes the code a little bit more straightforward.
Let me know what you think of it. 🙂

benchmarks/benches/songs.rs
Kerollmops: As you want but the real advantage of using the debug impl is that it also escape quotes and stuff like that :)
But, no we will keep it like that, don't change this line :)

benchmarks/benches/utils.rs
Kerollmops: Ho OK, couldn't you just use `drop`, instead? that's also a noop function.

Kerollmops (changes requested):

benchmarks/.gitignore
```diff
@@ -0,0 +1 @@
+benches/datasets_paths.rs
```
Kerollmops, line 1: Shouldn't we also ignore the `target` folder and the `Cargo.lock` here?

Kerollmops (changes requested):

benchmarks/.gitignore
irevoire: Since I added the `benchmarks` directory to the workspace defined in `milli/Cargo.toml` it shares it's `target` and `Cargo.lock` with `milli`, `http-ui`, etc in `milli/target` / `milli/Cargo.lock` (and these two are already ignored by `milli/.gitignore`

Kerollmops: Ho, yeah but why do we have a new gitignore here? Why don't we have this line in the current gitignore?

irevoire: I just thought it would be “easier” to maintain; If one day we decide to delete the `benchmarks` its `.gitignore` get deleted with it.

But honestly, it's as you want. I don't have any opinion on the subject.

Kerollmops: That's better indeed :)

Kerollmops (approved):
That's a very good PR and I am finally merging it, thank you @irevoire 🎉

bors merge

PR #210, opened 2021-06-03
Check the benchmarks in the CI

### LINKED ISSUE #209: Cargo check of benchmark in CI

Kerollmops:
Fixes #209.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,6 +37,11 @@ jobs:
         with:
           command: test
 
+      - uses: actions-rs/cargo@v1
+        with:
+          command: bench
+          args: --no-run -p benchmarks
+
       # - uses: actions-rs/cargo@v1
       #   with:
       #     command: bench
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,6 +31,11 @@ jobs:
 
       - uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -37,6 +37,11 @@ jobs:
         with:
           command: test
 
+      - uses: actions-rs/cargo@v1
+        with:
+          command: bench
+          args: --no-run -p benchmarks
+
```
ManyTheFish, line 44: can't we modify the cargo build action L32 by adding `args` under the `command` in order to check all packages?
```yaml
      - uses: actions-rs/cargo@v1
        with:
              command: build
              args: # ...
```

Kerollmops: Yup we could but add this too!

Kerollmops: I just added a `check --all` CI job.

ManyTheFish: If we already have a `cargo check --all` what is the purpose of an additional `cargo bench --no-run`?

Kerollmops: That's because `cargo bench` compiles the benchmarks which `cargo check` don't.

ManyTheFish (approved):

Kerollmops:
Thank you @ManyTheFish!
bors merge

PR #213, opened 2021-06-03
Fix the benchmarks script and names

Kerollmops:
The benchmarks compare script was not using the `--output` flag and was therefore failing the download of the JSON reports. We also modified the criterion benchmarks to use shorter names, it helps in looking at the benchmarks in the terminal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -900,7 +900,7 @@ dependencies = [
 
 [[package]]
 name = "helpers"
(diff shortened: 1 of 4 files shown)


curquiza (approved):
🔥

Kerollmops:
👊 bors merge

Kerollmops:
bors merge

PR #217, opened 2021-06-07
Improve the benchmarks readme

irevoire:
- Move the Dataset part to the end of the readme so when peoples just want to run the benchmarks they are not tempted to download the benchmarks by hand (which are going to be downloaded anyway by the `build.rs` script)
- Fix the links in the dataset -- wiki part


curquiza (changes requested):
You need to change the TOC order 🙂

curquiza (approved):
Checkd with @irevoire, the TOC change is already on `main`

Kerollmops (approved):
Thank you very much, merging!
bors merge

PR #218, opened 2021-06-08
Enable optimization for build.rs and macro

irevoire:
It fasten the unzip of the benchmark’s dataset a lot

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,3 +7,13 @@ debug = true
 
 [profile.bench]
 debug = true
+
+# Set the settings for build scripts and proc-macros.
+[profile.dev.build-override]
+opt-level = 3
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -7,3 +7,13 @@ debug = true
 
 [profile.bench]
 debug = true
+
+# Set the settings for build scripts and proc-macros.
```
Kerollmops, line 11:
```suggestion
# Make sure that the build scripts and proc-macros are compiled without
# all the optimizations. It speeds up the zip crate that we use in the build.rs.
```

Kerollmops (changes requested):

Kerollmops (approved):
Thank you!
bors merge

PR #219, opened 2021-06-08
Save the criteria field name in the fields ids map

Kerollmops:
We now make sure that we insert the new field name into the fields-ids-map when we are creating and modifying the criteria. Previously the engine was throwing an error due to that if the field was missing in the fields ids map.


ManyTheFish (dismissed):

Kerollmops:
bors merge

ManyTheFish (approved):

PR #222, opened 2021-06-08
Move the `UpdateStore` into the http-ui crate

Kerollmops:
We no more need to have the `UpdateStore` inside of the mill crate as this is the job of the caller to stack the updates and sequentially give them to milli.


ManyTheFish (dismissed):

ManyTheFish (approved):

Kerollmops (commented):
bors merge

PR #232, opened 2021-06-15
Format the whole project 

irevoire:


irevoire:
Actually, I'm not sure the special configuration @Kerollmops wanted is a good idea (`use_small_heuristics = "max"`).

It does this kind of thing to the code:

image
image

Please tell me what you think, but personally, I think the left version is really way more readable.
@Kerollmops, could you tell me again what this configuration was improving for you to see if we can enable only this part without breaking the iterators?

milli/src/search/mod.rs
```diff
@@ -152,14 +152,15 @@ impl<'a> Search<'a> {
...
+    ) -> anyhow::Result<SearchResult> {
         let mut offset = self.offset;
         let mut initial_candidates = RoaringBitmap::new();
         let mut excluded_candidates = RoaringBitmap::new();
         let mut documents_ids = Vec::with_capacity(self.limit);
 
-        while let Some(FinalResult { candidates, bucket_candidates, .. }) = criteria.next(&excluded_candidates)? {
+        while let Some(FinalResult { candidates, bucket_candidates, .. }) =
+            criteria.next(&excluded_candidates)?
+        {
```
Kerollmops, line 163: @irevoire here is an example of why I would like to keep this parameter, if it wasn't specified it would have resulted in something like that:

```
while let Some(FinalResult {
        candidates,
        bucket_candidates,
        ..
    }) = criteria.next(&excluded_candidates)? {
```

PR #224, opened 2021-06-09
Enable optimization in every profile

irevoire:

PR #234, opened 2021-06-16
Revert "Enable optimization in every profile"

### LINKED ISSUE #233: compiling tests in release takes too much time
We may revert #224

ManyTheFish:
compiling tests in release takes too much time.

Reverts meilisearch/milli#224

Fix #233

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,9 +8,6 @@ opt-level = 3
 [profile.release]
 debug = true
 
-[profile.test]
-opt-level = 3
-
 # Make sure that the build scripts and proc-macros are compiled with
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -2,14 +2,11 @@
 members = ["milli", "http-ui", "benchmarks", "infos", "helpers", "search"]
 default-members = ["milli"]
 
-[profile.dev]
-opt-level = 3
```
ManyTheFish, line 6: @irevoire, should we keep this?

irevoire: If it's not taking too much time, yes we should :slightly_smiling_face:

Kerollmops (approved):
bors merge

PR #237, opened 2021-06-17
change sub errors visibility

MarinPostma:
re-export sub-error types so they can be matched upon outside of milli.

### DIFF
--- a/milli/src/lib.rs
+++ b/milli/src/lib.rs
@@ -22,7 +22,7 @@ use fxhash::{FxHasher32, FxHasher64};
 use serde_json::{Map, Value};
 
 pub use self::criterion::{default_criteria, Criterion};
-pub use self::error::Error;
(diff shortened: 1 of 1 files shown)


milli/src/lib.rs
```diff
@@ -22,7 +22,7 @@ use fxhash::{FxHasher32, FxHasher64};
 use serde_json::{Map, Value};
 
 pub use self::criterion::{default_criteria, Criterion};
-pub use self::error::Error;
+pub use self::error::{Error, InternalError, UserError};
```
Kerollmops, line 25:
```suggestion
pub use self::error::{Error, FieldIdMapMissingEntry, InternalError, SerializationError, UserError};
```

Kerollmops (changes requested):
You can also re-export the other errors :)

Kerollmops (approved):
Seems fine to me!
bors merge

PR #240, opened 2021-06-17
Field distribution

### LINKED ISSUE #198: Fix stats of indexes
Using milli 0.2.1, I still have an issue with the stats route:
- [x] to stay iso, the fields in `fieldsDistribution` should be displayed in alphabetical order
- [x] the number is not good if I add twice the same documents -> the `fieldsDistribution` values are increased but in fact there is still the same number of documents

If possible, add tests

### LINKED ISSUE #199: fieldsDistribution is renamed into fieldDistribution
According to this issue (https://github.com/meilisearch/transplant/issues/188), `fieldsDistribution` is renamed into `fieldDistribution`.

Even if it does not impact directly the MeiliSearch users, you might want to stay consistent with MeiliSearch and remove this English mistake 😇

irevoire:
closes #199
closes #198

### DIFF
--- a/milli/src/update/index_documents/transform.rs
+++ b/milli/src/update/index_documents/transform.rs
@@ -25,7 +25,7 @@ const DEFAULT_PRIMARY_KEY_NAME: &str = "id";
 pub struct TransformOutput {
     pub primary_key: String,
     pub fields_ids_map: FieldsIdsMap,
-    pub fields_distribution: FieldsDistribution,
+    pub field_distribution: FieldsDistribution,
     pub external_documents_ids: ExternalDocumentsIds<'static>,
     pub new_documents_ids: RoaringBitmap,
     pub replaced_documents_ids: RoaringBitmap,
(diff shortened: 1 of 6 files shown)


milli/src/update/index_documents/transform.rs
```diff
@@ -419,18 +419,34 @@ impl Transform<'_, '_> {
...
+                    let base_obkv = self.index.documents.get(&self.rtxn, &key)?.ok_or(
+                        InternalError::DatabaseMissingEntry {
+                            db_name: db_name::DOCUMENTS,
+                            key: None,
+                        },
+                    )?;
+
+                    // we remove all the fields that were already counted
+                    for (field_id, _) in base_obkv.iter() {
+                        use std::collections::btree_map::Entry;
```
Kerollmops, line 432: Could you please import that but at the top of the file?

Kerollmops (changes requested):
Thank you very much for this bug fix, that was a long-standing one! 🎾

milli/src/update/index_documents/transform.rs
irevoire: done :slightly_smiling_face:

Kerollmops (approved):
Thank you very much Tamo 📽️ 
bors merge

PR #247, opened 2021-06-22
Return a `MissingDocumentId` error when a document doesn't have one

Kerollmops:
We were wrongly returning a `MissingPrimaryKey` instead of a `MissingDocumentId` error for when a document was missing a document id. We also improved the error message for when a document id is invalid (wrong type or wrong format).


curquiza (approved):

Kerollmops:
bors merge

PR #249, opened 2021-06-22
Enable the jemallocator dependencies only when we are running on linux

irevoire:


Kerollmops (changes requested):
Can you make sure that we correctly declare jemalloc as the global allocator, an only on Linux, in the main.rs files, please?

irevoire:
Yes I already checked :grin:

Kerollmops (approved):
Thank you very much!
bors merge

PR #251, opened 2021-06-22
Fix the limit

irevoire:
There was no check on the limit and thus if a user specified a very large number this line could cause a panic.


Kerollmops (approved):
That's true indeed, thank you for the fix!
bors merge

PR #257, opened 2021-06-23
Fix unconditional facet indexing

Kerollmops:
We were indexing every searchable field as filterable, this was a mistake.


curquiza (approved):

Kerollmops:
bors merge

PR #258, opened 2021-06-23
Replace the writer by a sorter when computing the level positions

Kerollmops:
This is an ugly hotfix, we had trouble understanding this issue and we need to fix it in a proper way.

I will let @curquiza create a tracking issue about the bug this PR fixes, IIRC, it is listed in the transplant repo.


curquiza:
Closed in favor of #261!

PR #264, opened 2021-06-28
Fix the invalid heed usage

### LINKED ISSUE #261: Unexpected key alterations when writing in LMDB can be caused by an undefined behavior
Hello,

I was trying to reproduce this bug https://github.com/meilisearch/transplant/issues/237, so I made a simple binary running only the necessary command to reproduce the bug, you can find it here: https://github.com/irevoire/milli_bug. I use this repository to run all of the commands below.

I also inserted a lot of debug print in heed (the LMDB Rust wrapper) so, as you can see in the `Cargo.toml`, I'm using my own version of heed that you can find here: https://github.com/irevoire/heed, on the "main" branch.

So basically, to summarize what I found, at some point, milli tries to insert an entry with a key equal to "and" by calling `heed::mdb::lmdb_ffi::mdb_cursor_put`, but in the end, the `&[u8]` milli gave to `mdb_cursor_put` got modified and contains "toi". I also find out that this "toi"  string is a view inside of the word "antoine", I didn't check but it is likely that it is in the same memory allocation.

It seems like the `Mdb_val` struct, which is the equivalent of a slice, i.e. a pointer and a length, is modified in a way that it moves the original pointer forward and makes us read memory at the wrong address, making us read "toi" instead of "and", but we keep the correct length. Also, I checked [this recent Reddit post about a similar bug](https://old.reddit.com/r/rust/comments/nqjyb7/the_most_annoying_bug_ive_had_to_track_down/), I tried to patch the [`RwCursor::put_current` method](https://github.com/irevoire/heed/blob/b93ff8d4b185e0984396e6778c03f0677957f3eb/heed/src/cursor.rs#L207-L232), by putting the `&mut` inside of the `unsafe` block but it didn't change anything, at least it didn't fix the bug.

@Kerollmops already tried to run `cargo miri` on the heed project but, unfortunately, as it is a wrapper on top of LMDB and `miri` doesn't check memory mapping correctness, it can't be checked with this great tool.

If you want to reproduce the bug, you can clone my [repository](https://github.com/irevoire/milli_bug) and run `cargo build`. Then run `rust-gdb target/debug/milli_bug` then:
 - Put a breakpoint in `b lmdb_ffi.rs:118`
 - Run the executable: `r`
 - Continue till the second indexation: `c`
 - Look at the variable named `key`: `x/s (*key).mv_data`, the value should be `and`.
 - Go just after the call to `mdb_cursor_put`: `n 5`
 - Look at the variable named `key`: `x/s key.data_ptr`, the value should be `toi`.

This is not supposed to be possible since LMDB does not modify this variable in place, at least that's what we think, but LMDB is a well-battle-tested library.

Kerollmops:
We were incorrectly using LMDB by keeping references into the database while modifying it, it was mostly due to heed not specifying anything about this behavior (blame me). I made sure in this PR that we always keep and **owned** version of the references we were previously using.

This PR partially fixes #261, we need to bump heed to v0.12 when it is available as it marks the `del/put_current`, and `append` methods as unsafe and explains why.

I would like to thank the whole Rust community for their help, especially:
 - @Pointerbender 🎸 who [tracked down the exact point where the memory write was altered](https://github.com/meilisearch/milli/issues/261#issuecomment-868983226),
 - @Diggsey 💯 who [found, in the LMDB documentation, the sentence that we forgot about](https://github.com/meilisearch/milli/issues/261#issuecomment-869002274),
 - and @RalfJung 🥇 who [track and kill undefined behavior in Rust programs in general](https://www.ralfj.de/).

### DIFF
--- a/milli/src/update/delete_documents.rs
+++ b/milli/src/update/delete_documents.rs
@@ -197,7 +197,8 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
                         iter.del_current()?;
                         *must_remove = true;
                     } else if docids.len() != previous_len {
-                        iter.put_current(key, &docids)?;
+                        let key = key.to_owned();
+                        iter.put_current(&key, &docids)?;
                     }
                 }
             }
@@ -238,13 +239,14 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
         let mut iter = word_prefix_docids.iter_mut(self.wtxn)?;
         while let Some(result) = iter.next() {
             let (prefix, mut docids) = result?;
+            let prefix = prefix.to_owned();
             let previous_len = docids.len();
             docids.difference_with(&self.documents_ids);
             if docids.is_empty() {
                 iter.del_current()?;
                 prefixes_to_delete.insert(prefix)?;
             } else if docids.len() != previous_len {
-                iter.put_current(prefix, &docids)?;
+                iter.put_current(&prefix, &docids)?;
             }
         }
 
@@ -281,7 +283,8 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
             if docids.is_empty() {
                 iter.del_current()?;
             } else if docids.len() != previous_len {
-                iter.put_current(key, &docids)?;
+                let key = key.to_owned();
+                iter.put_current(&key, &docids)?;
             }
         }
 
@@ -299,7 +302,8 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
             if docids.is_empty() {
                 iter.del_current()?;
             } else if docids.len() != previous_len {
-                iter.put_current(bytes, &docids)?;
+                let bytes = bytes.to_owned();
+                iter.put_current(&bytes, &docids)?;
             }
         }
 
@@ -315,7 +319,8 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
             if docids.is_empty() {
                 iter.del_current()?;
             } else if docids.len() != previous_len {
-                iter.put_current(bytes, &docids)?;
+                let bytes = bytes.to_owned();
+                iter.put_current(&bytes, &docids)?;
             }
         }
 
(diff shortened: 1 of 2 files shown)


irevoire (commented):
Tell me if I'm wrong, but we should totally fix these lines too, no?
https://github.com/meilisearch/milli/blob/main/milli/src/update/index_documents/mod.rs#L157
https://github.com/meilisearch/milli/blob/main/milli/src/update/index_documents/mod.rs#L219

Kerollmops:
The `k` variable comes from the grenad iterator, we compare it with the `key` variable which, this one, comes from the LMDB database. So, no need to change anything here, am I wrong?

RalfJung:
I obviously have no right to impose anything here, but I'd just suggest that when you make subtle changes like this, you leave a comment in the source code that at least briefly explains that this was a deliberate choice, and possibly links to an issue with more details. That reduces the risk of someone, in 2 years, going "ah I can eliminate some copies here" and re-introducing the bug. ;)

irevoire (approved):
Yes you're right, thank you!

And thanks again everyone :smile:

Kerollmops:
Thank you @RalfJung,

I am changing the heed functions to be unsafe, make a PR to this repo using the new version, and will add little comments above the functions in this PR. It will be just above an unsafe block, I hope this comment will be visible enough in 2 years 😄

Diggsey:
FWIW, you don't need to make the `heed` functions unsafe - you just need to make sure that all references returned from the database have a lifetime tied to the database, and that all database-modifying functions require a mutable borrow of the database in order to operate.

Kerollmops:
Thank you @Diggsey, I just opened an issue about your proposal, not sure if it will work but will try it out. I didn't publish the new v0.12 version of heed on crates.io so I can change anything.

Anyway, we will merge this PR for now as it fixes the bug and will use the potentially safe new version of heed if we update it!

Diggsey:
@Kerollmops just a heads up that without implementing my proposal, you might have to worry about race conditions, since one thread might be reading a key from the DB while another thread tries to write to the DB. (I'm not sure, maybe everything is already forced to be single-threaded)

Kerollmops:
@Diggsey, Yeah LMDB already ensures that there can be only one mutable transaction at a time. Thank you anyway!

MarinPostma:
I think that have to move around a ref mut to the db is too much constraint, since we won't be able to clone it and use references to it anymore. This is redundant to lmdb's locking mechanism.

As soon as we get GATs, which so not be [too long now](https://github.com/rust-lang/rust/issues/44265), we should be able to create a `Current` type which represent a view on the current iterator entry, with a consuming `self`, and accessors for the value and the data, whose lifetime are bound to the lifetime of the `Current` object.

PR #267, opened 2021-06-29
Highlighting

irevoire:
closes #262 
I basically rewrote a part of the damerau-levenshtein function we were using for the highlighting to accept at most two errors from the user and stop on the third mistake.
Also, now it supports utf-8, so it should fix our issue.

PR #272, opened 2021-06-30
Do not highlight too much when encountering prefixes

Kerollmops:
We were highlighting a little too much with the recent changes from #267, I fixed the `bytes_to_highlight` function to make it aware of the maximum number of typos to accept this way we only highlight the matching part according to the typo.


Kerollmops:
We found out that the problem was due to the query tree that was using an analyzer with a normalizing pipeline which was breaking the highlighting of emojis.

PR #275, opened 2021-07-01
Fix the benchmarks dependencies

irevoire:
Import exactly the same dependency as milli instead of a wildcard that can do anything

### DIFF
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -11,7 +11,7 @@ milli = { path = "../milli" }
 jemallocator = "0.3.2"
 
 [dev-dependencies]
-heed = { git = "https://github.com/Kerollmops/heed", tag = "v0.12.0", default-features = false, features = ["lmdb", "sync-read-txn"] }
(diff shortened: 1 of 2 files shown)


benchmarks/Cargo.toml
```diff
@@ -11,7 +11,7 @@ milli = { path = "../milli" }
 jemallocator = "0.3.2"
 
 [dev-dependencies]
-heed = "*" # we want to use the version milli uses
+heed = { git = "https://github.com/Kerollmops/heed", tag = "v0.12.0", default-features = false, features = ["lmdb", "sync-read-txn"] }
```
Kerollmops, line 14:
```suggestion
heed = { git = "https://github.com/Kerollmops/heed", tag = "v0.12.0" }
```

It is not mandatory to specify the features, IIRC.

Kerollmops (changes requested):

benchmarks/Cargo.toml
irevoire: That's right thanks :grinning:

Kerollmops (approved):
Thank you! 🥼 
bors merge

PR #276, opened 2021-07-01
Fix the fmt of the auto-generated file

irevoire:
The file generated by the `build.rs` file of the benchmark was badly formatted and that was causing an issue with the git pre-commit hook I wrote [earlier](https://github.com/meilisearch/milli/blob/main/script/pre-commit)


Kerollmops (approved):
Thank you! 🐤 
bors merge

PR #283, opened 2021-07-05
Use the AlwaysFreePages flag when opening an index

### LINKED ISSUE #279: Add LMDB flag to stop the memory "leak"

Kerollmops:
We introduced a new flag in our fork of LMDB, this `AlwaysFreePages` flag forces LMDB to always free the single pages it uses before writing to the disk instead of keeping them in a linked list.

Declaring this flag reduces the memory print (leak) we have on memory after indexing a lot of documents.

Fixes #279.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -873,7 +873,7 @@ dependencies = [
 [[package]]
 name = "heed"
 version = "0.12.0"
-source = "git+https://github.com/Kerollmops/heed?tag=v0.12.0#6c0b95793a805dc598f05c119494e6c069de0326"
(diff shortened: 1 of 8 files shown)


irevoire:
I think it works; I tried to execute a large indexation on my computer and looked at the memory usage of `http-ui` in `htop`. It was reporting a large memory usage but I suspect it's only cached memory that was freed but not reclaimed by the OS? :thinking: 

Whatever, I ran a second smaller indexation under `massif` and to me, the issue is fixed:
image
As you can see all the allocations related to `mdb_page_alloc` are well freed at the end of the indexation :tada:

irevoire (approved):
bors merge

Kerollmops:
@irevoire can you please do the same indexation but without this patch, please?

PR #284, opened 2021-07-05
[http-ui] Introduce the route `die`

irevoire:
This route just `exit` the process. This can come in handy when you run `http-ui` inside of another process (a profiler for example), and you don't want to kill everything

### DIFF
--- a/http-ui/src/main.rs
+++ b/http-ui/src/main.rs
@@ -983,7 +983,9 @@ async fn main() -> anyhow::Result<()> {
         });
 
     let die_route = warp::filters::method::get().and(warp::path!("die")).map(move || {
+        eprintln!("Killed by an HTTP request received on the die route");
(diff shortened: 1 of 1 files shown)


http-ui/src/main.rs
```diff
@@ -982,6 +982,11 @@ async fn main() -> anyhow::Result<()> {
             })
         });
 
+    let die_route = warp::filters::method::get().and(warp::path!("die")).map(move || {
+        std::process::exit(0);
```
Kerollmops, line 986:
```suggestion
        eprintln!("Killed by an HTTP request received on the die route");
        std::process::exit(0);
```

Kerollmops (changes requested):

Kerollmops (approved):
bors merge

PR #285, opened 2021-07-06
Support documents with at most 65536 fields

### LINKED ISSUE #248: Limit of fields
Currently, the limit of attributes in a doc is 256, this should be 65,536.

Kerollmops:
Fixes #248.

In this PR I updated the `obkv` crate, it now supports arbitrary key length and therefore I was able to use an `u16` to represent the fields instead of a single byte. It was impressively easy to update the whole codebase 🍡 🍔

### DIFF
--- a/infos/src/main.rs
+++ b/infos/src/main.rs
@@ -7,7 +7,7 @@ use byte_unit::Byte;
 use heed::EnvOpenOptions;
 use milli::facet::FacetType;
 use milli::index::db_name::*;
-use milli::{Index, TreeLevel};
+use milli::{FieldId, Index, TreeLevel};
 use structopt::StructOpt;
 use Command::*;
 
@@ -322,7 +322,7 @@ fn most_common_words(index: &Index, rtxn: &heed::RoTxn, limit: usize) -> anyhow:
 fn facet_values_iter<'txn, KC: 'txn, DC: 'txn>(
     rtxn: &'txn heed::RoTxn,
     db: heed::Database<KC, DC>,
-    field_id: u8,
+    field_id: FieldId,
 ) -> heed::Result<Box<dyn Iterator<Item = heed::Result<(KC::DItem, DC::DItem)>> + 'txn>>
 where
     KC: heed::BytesDecode<'txn>,
@@ -330,7 +330,7 @@ where
 {
     let iter = db
         .remap_key_type::<heed::types::ByteSlice>()
-        .prefix_iter(&rtxn, &[field_id])?
+        .prefix_iter(&rtxn, &field_id.to_be_bytes())?
         .remap_key_type::<KC>();
 
     Ok(Box::new(iter))
--- a/milli/src/heed_codec/facet/facet_level_value_f64_codec.rs
+++ b/milli/src/heed_codec/facet/facet_level_value_f64_codec.rs
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 use std::convert::TryInto;
 
 use crate::facet::value_encoding::f64_into_bytes;
-use crate::FieldId;
+use crate::{try_split_array_at, FieldId};
 
 // TODO do not de/serialize right bound when level = 0
 pub struct FacetLevelValueF64Codec;
@@ -11,7 +11,8 @@ impl<'a> heed::BytesDecode<'a> for FacetLevelValueF64Codec {
     type DItem = (FieldId, u8, f64, f64);
 
     fn bytes_decode(bytes: &'a [u8]) -> Option<Self::DItem> {
-        let (field_id, bytes) = bytes.split_first()?;
+        let (field_id_bytes, bytes) = try_split_array_at(bytes)?;
+        let field_id = u16::from_be_bytes(field_id_bytes);
         let (level, bytes) = bytes.split_first()?;
 
         let (left, right) = if *level != 0 {
@@ -23,7 +24,7 @@ impl<'a> heed::BytesDecode<'a> for FacetLevelValueF64Codec {
             (left, left)
         };
 
-        Some((*field_id, *level, left, right))
+        Some((field_id, *level, left, right))
     }
 }
 
@@ -61,8 +62,8 @@ impl heed::BytesEncode<'_> for FacetLevelValueF64Codec {
             16 // length
         };
 
-        let mut bytes = Vec::with_capacity(len + 2);
-        bytes.push(*field_id);
+        let mut bytes = Vec::with_capacity(len + 3);
+        bytes.extend_from_slice(&field_id.to_be_bytes());
         bytes.push(*level);
         bytes.extend_from_slice(&buffer[..len]);
         Some(Cow::Owned(bytes))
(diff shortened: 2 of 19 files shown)


milli/src/heed_codec/facet/facet_level_value_f64_codec.rs
```diff
@@ -61,8 +62,8 @@ impl heed::BytesEncode<'_> for FacetLevelValueF64Codec {
             16 // length
         };
 
-        let mut bytes = Vec::with_capacity(len + 2);
-        bytes.push(*field_id);
+        let mut bytes = Vec::with_capacity(len + 3);
```
irevoire, line 65:
```suggestion
        let mut bytes = Vec::with_capacity(len + 1 + mem::size_of::<FieldId>());
```

milli/src/heed_codec/facet/facet_value_string_codec.rs
```diff
@@ -1,14 +1,14 @@
...
-use crate::FieldId;
+use crate::{try_split_array_at, FieldId};
 
 pub struct FacetValueStringCodec;
 
 impl FacetValueStringCodec {
     pub fn serialize_into(field_id: FieldId, value: &str, out: &mut Vec<u8>) {
-        out.reserve(value.len() + 1);
-        out.push(field_id);
+        out.reserve(value.len() + 2);
```
irevoire, line 10:
```suggestion
        out.reserve(value.len() + mem::size_of::<FieldId>());
```

milli/src/heed_codec/facet/field_doc_id_facet_f64_codec.rs
```diff
@@ -2,31 +2,32 @@ use std::borrow::Cow;
...
     }
 }
 
 impl<'a> heed::BytesEncode<'a> for FieldDocIdFacetF64Codec {
     type EItem = (FieldId, DocumentId, f64);
 
     fn bytes_encode((field_id, document_id, value): &Self::EItem) -> Option<Cow<[u8]>> {
-        let mut bytes = Vec::with_capacity(1 + 4 + 8 + 8);
-        bytes.push(*field_id);
+        let mut bytes = Vec::with_capacity(2 + 4 + 8 + 8);
```
irevoire, line 29:
```suggestion
        let mut bytes = Vec::with_capacity(mem::size_of::<FieldId>() + 4 + 8 + 8);
```

milli/src/heed_codec/field_id_word_count_codec.rs
```diff
@@ -18,6 +19,9 @@ impl<'a> heed::BytesEncode<'a> for FieldIdWordCountCodec {
     type EItem = (FieldId, u8);
 
     fn bytes_encode((field_id, word_count): &Self::EItem) -> Option<Cow<[u8]>> {
-        Some(Cow::Owned(vec![*field_id, *word_count]))
+        let mut bytes = Vec::with_capacity(2 + 1);
```
irevoire, line 22:
```suggestion
        let mut bytes = Vec::with_capacity(mem::size_of::<FieldId>() + 1);
```

milli/src/index.rs
```diff
@@ -523,10 +523,11 @@ impl Index {
         field_id: FieldId,
         docids: &RoaringBitmap,
     ) -> heed::Result<()> {
-        let mut buffer = [0u8; main_key::STRING_FACETED_DOCUMENTS_IDS_PREFIX.len() + 1];
+        let mut buffer = [0u8; main_key::STRING_FACETED_DOCUMENTS_IDS_PREFIX.len() + 2];
```
irevoire, line 526:
```suggestion
        let mut buffer = [0u8; main_key::STRING_FACETED_DOCUMENTS_IDS_PREFIX.len() + mem::size_of::<FieldId>()];
```

milli/src/index.rs
```diff
@@ -536,10 +537,11 @@ impl Index {
         rtxn: &RoTxn,
         field_id: FieldId,
     ) -> heed::Result<RoaringBitmap> {
-        let mut buffer = [0u8; main_key::STRING_FACETED_DOCUMENTS_IDS_PREFIX.len() + 1];
+        let mut buffer = [0u8; main_key::STRING_FACETED_DOCUMENTS_IDS_PREFIX.len() + 2];
```
irevoire, line 540:
```suggestion
        let mut buffer = [0u8; main_key::STRING_FACETED_DOCUMENTS_IDS_PREFIX.len() + mem::size_of::<FieldId>()];
```

milli/src/index.rs
```diff
@@ -553,10 +555,11 @@ impl Index {
         field_id: FieldId,
         docids: &RoaringBitmap,
     ) -> heed::Result<()> {
-        let mut buffer = [0u8; main_key::NUMBER_FACETED_DOCUMENTS_IDS_PREFIX.len() + 1];
+        let mut buffer = [0u8; main_key::NUMBER_FACETED_DOCUMENTS_IDS_PREFIX.len() + 2];
```
irevoire, line 558:
```suggestion
        let mut buffer = [0u8; main_key::NUMBER_FACETED_DOCUMENTS_IDS_PREFIX.len() + mem::size_of::<FieldId>()];
```

infos/src/main.rs
```diff
@@ -7,7 +7,7 @@ use byte_unit::Byte;
 use heed::EnvOpenOptions;
 use milli::facet::FacetType;
 use milli::index::db_name::*;
-use milli::{Index, TreeLevel};
+use milli::{FieldId, Index, TreeLevel};
 use structopt::StructOpt;
 use Command::*;
```
irevoire, line 13:
```suggestion
use std::mem;
```

irevoire (dismissed):
I'm mostly ok with the code but I think we should remove as many “magic numbers” as possible from the codebase.
Here I think we could pretty easily use the function `size_of` to be more expressive about what is happening and having less code to refactor if one day we need to change `FieldId` again.
Personally, to me the code is easier to read this way and should be easier to maintain, let me know what you think of this change.

My other concern was about the tests, maybe we could add a test that tries to add too many `fields` and check we really get the error we are expecting?

irevoire (approved):

Kerollmops:
Thank you very much, merging!
bors merge

MarinPostma:
jumping on @irevoire suggestion: I understand that it is less compact to use those method, but using harcoded values for values that are semantacally tied to the size of a type is a code smell. It's a matter of time until these things come back biting at us. Hopefully you haven't forgotten any of those number ;)

irevoire:
Even if **we** don't forget these numbers I'm afraid of contributors coming and seeing things like that:
```
let mut bytes = Vec::with_capacity(2 + 4 + 8 + 8); // why is there 4 numbers when an EItem only contains 3 members? 

let mut bytes = Vec::with_capacity(len + 3); // why is that a 3 already?
```
 And talking for myself, I'm pretty sure I'll forget all of that before Monday :joy:

Kerollmops:
@MarinPostma Just don't forget that it is only used for `Vec::with_capacity` or `Vec::reserve`, it can only reduce the performance of the Vec creation, also note that the first allocation of a Vec is 8 bytes IIRC. The first allocation already cover ours.

Kerollmops:
bors merge

PR #288, opened 2021-07-07
Stop tracking the Cargo.lock and add cache + windows to the CI

irevoire:
We reuse the same `~/.cargo` and `./target` directory between each run on the same OS and rust toolchain.
The `key` to decide if we can use the cache or not is: `$OS_NAME-$RUST_TOOLCHAIN-$HASH(Cargo.toml)`

We also removed the `Cargo.lock` from this repository. Indeed, milli is a library and [should not track the `Cargo.lock`](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html)

And finally, we enabled the tests on `windows-latest`. Since `lmdb` has been updated, this is now possible.


irevoire:
I had no luck on the last run, it took a really long time to find a macOS workers. But apart from that, the CI look really faster, I think we can merge that.

Since I'm on it do you think we should run the tests on windows now? (it works on transplant so it should also works on milli)

Kerollmops (changes requested):
Indeed, it would be great to test on Windows.
Could you also remove the Cargo.lock, add it to the .gitignore and patch the CI to not check with `--locked`?

Kerollmops (approved):
Seems good to me, thank you!
bors merge

Kerollmops:
bors merge

irevoire:
bors cancel

Kerollmops (approved):
bors merge

irevoire:
bors merge

Kerollmops:
bors merge

Kerollmops:
bors merge

curquiza:
It should work now

curquiza:
bors merge

curquiza (approved):

PR #290, opened 2021-07-19
Add a $HOME to the CI

irevoire:
This should fix this issue:
https://github.com/meilisearch/milli/runs/3104228432?check_suite_focus=true

I think a real fix would be to fix the configuration of our github runner but I don't know how to do it.
@curquiza could probably help us on that once she's back from vacation :smile:


Kerollmops (approved):
Thank you, I hope this fix works!
bors merge

PR #302, opened 2021-08-04
Update milli to v0.9.0

curquiza:
Updating the minor and not patch since #300 seems to be breaking: it involves a re-indexation to get the fix, so it involves an additional step from the users, not only downloading the latest version.


ManyTheFish (approved):

curquiza:
bors merge

curquiza:
bors merge

PR #303, opened 2021-08-12
Remove max values by facet limit for facet distribution

ManyTheFish:


curquiza (approved):

Kerollmops (approved):
Seems good to me!
bors merge

bb:
Regarding the **breaking-change** label: I think this is breaking compared to earlier 0.21.x versions but it's in line with <=0.20.x, isn't it?

curquiza:
This is breaking for the users of the library Milli, so for the MeiliSearch repository. But a breaking on Milli's side does not involve this is breaking for MeiliSearch.
Here, in this PR, a public function changed, so the milli's library is breaking.

PR #306, opened 2021-08-13
Update grenad to get better assertion debug info

bb:
https://meilicommunity.slack.com/archives/CP9DVS1RQ/p1628858601125800?thread_ts=1626768225.066500&cid=CP9DVS1RQ


bb:
Only upgrading the version is not enough due to API changes, so this PR is useless. The FuseFile stuff is easily done but the ChunkCreator changes drag on quite a bit into Sorter and the indexes (word prefix, pair proximity, word level positions). It's better be done by the core team.

PR #309, opened 2021-08-17
Sort at query time

### LINKED ISSUE #305: Sort at the query time
### Important

- This issue is based on this specification: https://github.com/meilisearch/specifications/blob/main/text/0055-sort.md
The specification gathers all the information related to this feature: the motivation, the API and technical changes, the future possibilities...
⚠️ For anything that is not clear or deeply detailed in this issue, please refer to this specification.

- This issue is a sub-issue of the tracking issue on MeiliSearch's side: https://github.com/meilisearch/MeiliSearch/issues/1581

---

## Feature purpose

The sort feature introduces the possibility to quickly sort the search results as an end-user during the search time.

## TODO

- [ ] The core engine should support the sort on `String` in lexicographic order. If the dataset contains values in `String` and `number` for the same field, the search engine first sorts the numbers and then the strings.
`asc`/`desc` should now work on strings with the same behavior

- [ ] Add the `sort` ranking rule at the third position by default. It means the core-engine will apply the sort after the ranking rules `words` and `typo`. If the end-user does not request any search at the query time, the ranking rule `sort` is ignored.
The `sort` ranking rule can be removed or set at a different place in the `rankingRules` setting, so that the users can adjust their relevancy.

- [ ] Add `sort` query/request parameter during search time. It supports strings and numbers. Strings can be sorted in lexicographical order.
The end-users must indicate the order of the expected sort (asc/desc). Multiple fields can be passed to be sorted:
ex: `["price:asc", "author:desc"]`
The error message when the end-users try to sort on an attributes non-declared as `sortableAttributes`:

```
Attribute `title` is not sortable, available sortable attributes are: ..., ...
```

- [ ] The search engine needs to know what are the attributes to sort via `sortableAttributes` settings.
The behavior of this setting is similar to `filterableAttributes`: the default value is `[]` and the core engine expects an array of strings. If an unknown field is passed to `sortableAttributes`, no error is raised.
Refer to the spec for more details and the error behaviors.

- [ ] New syntax for the custom ranking rules: `desc(X)` becomes `X:desc` and `asc(Y)` becomes `Y:asc`. This syntax is also available for sorting during query time.

- [ ] Add tests for the sort relevancy

Kerollmops:
This PR:
 - Makes the `Asc/Desc` criteria work with strings too, it first returns documents ordered by numbers then by strings, and finally the documents that can't be ordered. Note that it is lexicographically ordered and not ordered by character, which means that it doesn't know about wide and short characters i.e. `a`, `丹`, `▲`.
 - Changes the syntax for the `Asc/Desc` criterion by now using a colon to separate the name and the order i.e. `title:asc`, `price:desc`.
 - Add the `Sort` criterion at the third position in the ranking rules by default.
 - Add the `sort_criteria` method to the `Search` builder struct to let the users define the `Asc/Desc` sortable attributes they want to use at query time. Note that we need to check that the fields are registered in the sortable attributes before performing the search.
 - Introduce a new `InvalidSortableAttribute` user error that is raised when the sort criteria declared at query time are not part of the sortable attributes.
 - @ManyTheFish introduced integration tests for the dynamic Sort criterion.

Fixes #305.


ManyTheFish (approved):

ManyTheFish (approved):

Kerollmops:
bors merge

PR #315, opened 2021-08-19
Rewrite the indexing benchmarks

irevoire:
There was a panic on the benchmark and while I was trying to understand what was happening I decided to rewrite the way the benchmarks were working.

Before we were creating a database with the good setting, and then for each benchmarks we were:
1. Deleting all documents in the database
2. Indexing a batch of documents

Now for each iteration we recreate entirely a new database from scratch.
Since deleting all the documents in a database may not be the same as starting with a fresh new database I prefer this solution.

### DIFF
--- a/benchmarks/benches/indexing.rs
+++ b/benchmarks/benches/indexing.rs
@@ -31,48 +31,43 @@ fn setup_index() -> Index {
 }
 
 fn indexing_songs_default(c: &mut Criterion) {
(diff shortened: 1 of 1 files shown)


irevoire:
I ran the full bench suit on the server and it works so I think we can merge 🎉

Kerollmops (approved):
THank you a lot for this, LGTM!
bors merge

PR #317, opened 2021-08-20
Fix the facet string docids filterable deletion bug

Kerollmops:
Fixes a bug where the deletion of documents was returning a decoding error. But only when the settings are set with filterable attributes.

This bug was introduced in #254 in which we made the engine faster in returning the facet distribution. We changed the way we were storing the inverted index, we were no more storing only documents ids with the original values but also groups identified with integers, depending on the facet level we were using. This is similar to how facet numbers are already stored.

⚠️ As @curquiza already said, we must first revert #309 before merging this!

Related to https://github.com/meilisearch/MeiliSearch/issues/1601.

### DIFF
--- a/milli/src/update/delete_documents.rs
+++ b/milli/src/update/delete_documents.rs
@@ -4,12 +4,15 @@ use std::collections::HashMap;
 use chrono::Utc;
 use fst::IntoStreamer;
 use heed::types::ByteSlice;
+use heed::{BytesDecode, BytesEncode};
 use roaring::RoaringBitmap;
 use serde_json::Value;
 
 use super::ClearDocuments;
-use crate::error::{InternalError, UserError};
-use crate::heed_codec::facet::FacetStringLevelZeroValueCodec;
+use crate::error::{InternalError, SerializationError, UserError};
+use crate::heed_codec::facet::{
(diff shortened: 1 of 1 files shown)


curquiza (dismissed):
I tested with MeiliSearch, it works 🎉

milli/src/update/delete_documents.rs
```diff
@@ -566,4 +606,51 @@ mod tests {
...
+        let mut builder = IndexDocuments::new(&mut wtxn, &index, 0);
+        builder.update_format(UpdateFormat::Json);
+        builder.execute(content, |_, _| ()).unwrap();
+
+        // Delete not all of the documents but some of them.
+        let mut builder = DeleteDocuments::new(&mut wtxn, &index, 1).unwrap();
+        builder.delete_external_id("1_4");
+        builder.execute().unwrap();
+
+        wtxn.commit().unwrap();
```
ManyTheFish, line 654: Why don't we check that the corresponding ID is deleted from the facet database or searching for the label `drawing` and checking if the document `1_4` is not present?

ManyTheFish (changes requested):

curquiza (approved):

curquiza (approved):

ManyTheFish (approved):

Kerollmops:
bors merge

PR #321, opened 2021-08-23
Run the benchmarks once every weeks in the CI

### LINKED ISSUE #314: Benchmark CI in a cron
To anticipate benchmark failures we should add a cron CI with GitHub action that run the benchmark on `main` every week 🙂

irevoire:
closes #314


irevoire:
Avoiding code duplication was too hard so we decided to copy and paste the original `benchmark.yaml` file once for every benchmark.
You can see all my unsuccessful attempts in the commits of this PR.
And this PR is closed in favor of #329

PR #322, opened 2021-08-23
Geosearch

irevoire:
This PR introduces [basic geo-search functionalities](https://github.com/meilisearch/specifications/pull/59), it makes the engine able to index, filter and, sort by geo-point. We decided to use [the rstar library](https://docs.rs/rstar) and to save the points in [an RTree](https://docs.rs/rstar/0.9.1/rstar/struct.RTree.html) that we de/serialize in the index database [by using serde](https://serde.rs/) with [bincode](https://docs.rs/bincode). This is not an efficient way to query this tree as it will consume a lot of CPU and memory when a search is made, but at least it is an easy first way to do so.

### What we will have to do on the indexing part:
 - [x] Index the `_geo` fields from the documents.
   - [x] Create a new module with an extractor in the `extract` module that takes the `obkv_documents` and retrieves the latitude and longitude coordinates, outputting them in a `grenad::Reader` for further process.
   - [x] Call the extractor in the `extract::extract_documents_data` function and send the result to the `TypedChunk` module.
   - [x] Get the `grenad::Reader` in the `typed_chunk::write_typed_chunk_into_index` function and store all the points in the `rtree`
- [x] Delete the documents from the `RTree` when deleting documents from the database. All this can be done in the `delete_documents.rs` file by getting the data structure and removing the points from it, inserting it back after the modification.
- [x] Clearing the `RTree` entirely when we clear the documents from the database, everything happens in the `clear_documents.rs` file.
- [x] save a Roaring bitmap of all documents containing the `_geo` field

### What we will have to do on the query part:
- [x] Filter the documents at a certain distance around a point, this is done by [collecting the documents from the searched point](https://docs.rs/rstar/0.9.1/rstar/struct.RTree.html#method.nearest_neighbor_iter) while they are in range.
  - [x] We must introduce new `geoLowerThan` and `geoGreaterThan` variants to the `Operator` filter enum.
  - [x] Implement the `negative` method on both variants where the `geoGreaterThan` variant is implemented by executing the `geoLowerThan` and removing the results found from the whole list of geo faceted documents.
  - [x] Add the `_geoRadius` function in the pest parser.
- [x] Introduce a `_geo` ascending ranking function that takes a point in parameter, ~~this function must keep the iterator on the `RTree` and make it peekable~~ This was not possible for now, we had to collect the whole iterator. Only the documents that are part of the candidates must be sent too!
  - [x] This ascending ranking rule will only be active if the search is set up with the `_geoPoint` parameter that indicates the center point of the ascending ranking rule.

-----------

- On Meilisearch part: We must introduce a new concept, returning the documents with a new `_geoDistance` field when it passed by the `_geo` ranking rule, this has never been done before. We could maybe just do it afterward when the documents have been retrieved from the database, computing the distance from the `_geoPoint` and all of the documents to be returned.


milli/src/update/index_documents/typed_chunk.rs
```diff
@@ -176,6 +178,22 @@ pub(crate) fn write_typed_chunk_into_index(
             )?;
             is_merged_database = true;
         }
+        TypedChunk::GeoPoints(mut geo_points) => {
+            // TODO: TAMO: we should create the rtree with the `RTree::bulk_load` function
+            let mut rtree = index.geo_rtree(&index.read_txn()?)?.unwrap_or_default();
```
Kerollmops, line 183:
```suggestion
            let mut rtree = index.geo_rtree(&wtxn)?.unwrap_or_default();
```

You must always read your database with thee same, with the same transaction. use the `wtxn` given to you.

Kerollmops (changes requested):

milli/src/update/index_documents/typed_chunk.rs
```diff
@@ -176,6 +178,22 @@ pub(crate) fn write_typed_chunk_into_index(
...
+            let mut rtree = index.geo_rtree(wtxn)?.unwrap_or_default();
+            while let Some((key, value)) = geo_points.next()? {
+                // convert the key back to a u32 (4 bytes)
+                let (key, _) = helpers::try_split_array_at::<u8, 4>(key).unwrap();
+                let key = u32::from_le_bytes(key);
+
+                // convert the latitude and longitude back to a f64 (8 bytes)
+                let (lat, tail) = helpers::try_split_array_at::<u8, 8>(value).unwrap();
+                let (lng, _) = helpers::try_split_array_at::<u8, 8>(tail).unwrap();
+                let point = [f64::from_le_bytes(lat), f64::from_le_bytes(lng)];
```
Kerollmops, line 192:
```suggestion
                let key = u32::from_be_bytes(key);

                // convert the latitude and longitude back to a f64 (8 bytes)
                let (lat, tail) = helpers::try_split_array_at::<u8, 8>(value).unwrap();
                let (lng, _) = helpers::try_split_array_at::<u8, 8>(tail).unwrap();
                let point = [f64::from_ne_bytes(lat), f64::from_ne_bytes(lng)];
```

Here you must also read native endian from the value bytes. And the docid is encoded in big-endian.

Kerollmops (changes requested):

milli/src/update/index_documents/extract/extract_geo_points.rs
```diff
@@ -0,0 +1,46 @@
...
+    let geo_field_id = geo_field_id.unwrap();
+
+    while let Some((docid_bytes, value)) = obkv_documents.next()? {
+        let obkv = obkv::KvReader::new(value);
+        let point = obkv.get(geo_field_id).unwrap(); // TODO: TAMO where should we handle this error?
+        let point: Value = serde_json::from_slice(point).map_err(InternalError::SerdeJson)?;
+
+        if let Some((lat, lng)) = point["lat"].as_f64().zip(point["lng"].as_f64()) {
+            // this will create an array of 16 bytes (two 8 bytes floats)
+            let bytes: [u8; 16] = concat_arrays![lat.to_le_bytes(), lng.to_le_bytes()];
```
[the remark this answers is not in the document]
irevoire: oops yes that’s what I wanted to do! Good catch thank you

milli/src/update/index_documents/typed_chunk.rs
irevoire: done thanks 👍

milli/src/update/index_documents/extract/extract_geo_points.rs
```diff
@@ -0,0 +1,46 @@
...
+) -> Result<grenad::Reader<File>> {
+    let mut writer = tempfile::tempfile().and_then(|file| {
+        create_writer(indexer.chunk_compression_type, indexer.chunk_compression_level, file)
+    })?;
+
+    // we never encountered any documents with a `_geo` field. We can skip entirely this step
+    if geo_field_id.is_none() {
+        return Ok(writer_into_reader(writer)?);
+    }
+    let geo_field_id = geo_field_id.unwrap();
```
Kerollmops, line 27:
```suggestion
    let geo_field_id = match geo_field_id {
        Some(geo) => geo,
        None => return Ok(writer_into_reader(writer)?),
    };
```

Kerollmops (changes requested):

milli/src/update/index_documents/extract/mod.rs
```diff
@@ -181,7 +181,7 @@ fn extract_documents_data(
     faceted_fields: &HashSet<FieldId>,
     stop_words: &Option<fst::Set<&[u8]>>,
 ) -> Result<(
-    grenad::Reader<CursorClonableMmap>,
+    (grenad::Reader<CursorClonableMmap>, grenad::Reader<CursorClonableMmap>),
```
ManyTheFish, line 184:
```rust
    grenad::Reader<CursorClonableMmap>, (grenad::Reader<CursorClonableMmap>,
    (grenad::Reader<CursorClonableMmap>, grenad::Reader<CursorClonableMmap>)),
```

milli/src/update/index_documents/extract/extract_geo_points.rs
```diff
@@ -0,0 +1,49 @@
...
+use super::helpers::{create_writer, writer_into_reader, GrenadParameters};
+use crate::{FieldId, InternalError, Result};
+
+/// Extracts the geographical coordinates contained in each document under the `_geo` field.
+///
+/// Returns the generated grenad reader containing the docid as key associated to the (latitude, longitude)
+pub fn extract_geo_points<R: io::Read>(
+    mut obkv_documents: grenad::Reader<R>,
+    indexer: GrenadParameters,
+    geo_field_id: Option<FieldId>, // faire un grenad vide
```
ManyTheFish, line 17: why don't you call `extract_geo_points` only if a geo_field_id exists?

milli/src/update/index_documents/extract/extract_geo_points.rs
```diff
@@ -0,0 +1,49 @@
...
+        };
+        let point: Value = serde_json::from_slice(point).map_err(InternalError::SerdeJson)?;
+
+        if let Some((lat, lng)) = point["lat"].as_f64().zip(point["lng"].as_f64()) {
+            // this will create an array of 16 bytes (two 8 bytes floats)
+            let bytes: [u8; 16] = concat_arrays![lat.to_ne_bytes(), lng.to_ne_bytes()];
+            writer.insert(docid_bytes, bytes)?;
+        } else {
+            // TAMO: improve the warn
+            warn!("Malformed `_geo` field");
```
ManyTheFish, line 43: we may return an error. For the moment any error in a document stop the indexation

milli/src/update/index_documents/extract/mod.rs
```diff
@@ -177,6 +182,15 @@ fn extract_documents_data(
...
     let _ = lmdb_writer_sx.send(Ok(TypedChunk::Documents(documents_chunk.clone())));
 
+    let documents_chunk_cloned = documents_chunk.clone();
+    let lmdb_writer_sx_cloned = lmdb_writer_sx.clone();
+    rayon::spawn(move || {
+        let _ = match extract_geo_points(documents_chunk_cloned, indexer, geo_field_id) {
+            Ok(geo_points) => lmdb_writer_sx_cloned.send(Ok(TypedChunk::GeoPoints(geo_points))),
+            Err(error) => lmdb_writer_sx_cloned.send(Err(error)),
+        };
+    });
```
ManyTheFish, line 192:
```suggestion
    if let Some(geo_field_id) = geo_field_id {
        let documents_chunk_cloned = documents_chunk.clone();
        let lmdb_writer_sx_cloned = lmdb_writer_sx.clone();
        rayon::spawn(move || {
            let _ = match extract_geo_points(documents_chunk_cloned, indexer, geo_field_id) {
                Ok(geo_points) => lmdb_writer_sx_cloned.send(Ok(TypedChunk::GeoPoints(geo_points))),
                Err(error) => lmdb_writer_sx_cloned.send(Err(error)),
            };
        });
    }
```

ManyTheFish: this will remove your Option in `extract_geo_points`

ManyTheFish:
Don't forget the document deletion 🤓

ManyTheFish (changes requested):

milli/src/update/index_documents/extract/extract_geo_points.rs
irevoire: Yep that's smarter actually :joy:

milli/src/update/index_documents/extract/extract_geo_points.rs
irevoire: Are we sure we want to stop the whole indexation for one malformed document? :thinking: 
@Kerollmops to do a simple warn I think, so I went with it

Kerollmops: I summon @gmourier for this, but I highly recommend not stop the whole indexation but rather just log that the document with external id `X` has an incorrect `_geo` field format. You can see how I have done this in the `docid_word_positions.rs` where we log the words that are too long (checked with the `valid_mdb_key` function).

ManyTheFish: My POV is: "a transaction is committed only if everything is ok. There is no partially valid transaction".
But in another way, the facets fields that don't have a good format are just ignored without printing any warning. 🤷

irevoire:
We had a debate with @ManyTheFish about the `AscDesc` enum in `scr/criterion.rs`.
I like the version I wrote, but if I remember correctly @ManyTheFish  was worried that it would add too much useless pattern match and proposed to use an enum looking like that:
```rust
pub enum AscDesc {
    Asc(String),
    Desc(String),
    GeoAsc([f64; 2]),
    GeoDesc([f64; 2]),
}
```
Now that we can see what is the total implication of embedding an enum in the `AscDesc` enum let me know what you think of it. 
Personally, I like it more this way, but I'll follow the majority! :grin:

milli/src/error.rs
```diff
@@ -12,6 +12,12 @@ use crate::{DocumentId, FieldId};
 
 pub type Object = Map<String, Value>;
 
+const RESERVED_KEYWORD: &[&'static str] = &["_geo", "_geoDistance"];
+
+pub fn is_reserved_keyword(keyword: &str) -> bool {
+    RESERVED_KEYWORD.contains(&keyword)
+}
```
Kerollmops, line 19:
```suggestion
pub fn is_reserved_keyword(keyword: &str) -> bool {
    ["_geo", "_geoDistance"].contains(&keyword)
}
```

milli/src/search/facet/filter_condition.rs
```diff
@@ -156,6 +161,65 @@ impl FilterCondition {
...
+            .map_err(UserError::InvalidFilter)?;
+        if parameters.len() != 3 {
+            return Err(UserError::InvalidFilter(PestError::new_from_span(
+                        ErrorVariant::CustomError {
+                            message: format!("The `_geoRadius` filter expect three arguments: `_geoRadius(latitude, longitude, radius)`"),
+                        },
+                    // we want to point to the last parameters and if there was no parameters we
+                        // point to the parenthesis
+                    parameters.last().map(|param| param.1.clone()).unwrap_or(param_span),
+                            )))?;
```
Kerollmops, line 204: Are you sure this part is correctly formatted? Seems strange to me.

milli/src/search/facet/filter_condition.rs
```diff
@@ -505,6 +598,18 @@ fn field_id(
 ) -> StdResult<Option<FieldId>, PestError<Rule>> {
     // lexing ensures that we at least have a key
     let key = items.next().unwrap();
+    if key.as_rule() == Rule::reserved {
+        return Err(PestError::new_from_span(
+            ErrorVariant::CustomError {
+                message: format!(
+                    "`{}` is a reserved keyword and thus can't be used as a filter expression. Available filterable attributes are: {}",
```
Kerollmops, line 605:
```suggestion
                    "`{}` is a reserved keyword and therefore can't be used as a filter expression. \
                    Available filterable attributes are: {}",
```

milli/src/search/facet/filter_condition.rs
```diff
@@ -663,6 +775,92 @@ mod tests {
...
+            )),
+            Box::new(Operator(1, LowerThanOrEqual(10.))),
+        );
+        assert_eq!(condition, expected);
+
+        // georadius don't have any parameters
+        let result = FilterCondition::from_str(&rtxn, &index, "_geoRadius");
+        assert!(result.is_err());
+        let error = result.unwrap_err();
+        assert!(error.to_string().contains("The `_geoRadius` filter expect three arguments: `_geoRadius(latitude, longitude, radius)`"));
```
Kerollmops, line 826: @gmourier We must not forget to specify somewhere that the radius is in meters.

milli/src/search/facet/grammar.pest
```diff
@@ -8,6 +8,9 @@ char =  _{ !(PEEK | "\\") ~ ANY
     | "\\" ~ (PEEK | "\\" | "/" | "b" | "f" | "n" | "r" | "t")
     | "\\" ~ ("u" ~ ASCII_HEX_DIGIT{4})}
 
+reserved = { "_geo" | "_geoDistance" | "_geoPoint" | ("_geoPoint" ~ parameters) }
+// we deliberately choose to allow empty parameters to generate more specific error message later
```
Kerollmops, line 12: Maybe using nom as the parser would help increase this, instead of unconstraining the pest parser to catch errors on the Rust side.

milli/src/search/mod.rs
```diff
@@ -148,13 +148,15 @@ impl<'a> Search<'a> {
         if let Some(sort_criteria) = &self.sort_criteria {
             let sortable_fields = self.index.sortable_fields(self.rtxn)?;
             for asc_desc in sort_criteria {
-                let field = asc_desc.field();
-                if !sortable_fields.contains(field) {
-                    return Err(UserError::InvalidSortableAttribute {
-                        field: field.to_string(),
-                        valid_fields: sortable_fields,
+                // we are not supposed to find any geoPoint in the criterion
+                if let Some(field) = asc_desc.field() {
```
Kerollmops, line 152: Ok, so this line explains why we match an optional here, as the `_geo` criterion does not have a field value (like for `price` or `title`).

milli/src/update/delete_documents.rs
```diff
@@ -380,6 +380,23 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
...
+
+            let points_to_remove: Vec<_> = rtree
+                .iter()
+                .filter(|&point| self.documents_ids.contains(point.data))
+                .cloned()
+                .collect();
+            points_to_remove.iter().for_each(|point| {
+                rtree.remove(&point);
+                geo_faceted_doc_ids.remove(point.data);
+            });
```
Kerollmops, line 394:
```suggestion
            let (points_to_remove, docids_to_remove): (Vec<_>, RoaringBitmap) = rtree
                .iter()
                .filter(|&point| self.documents_ids.contains(point.data))
                .cloned()
                .map(|point| (point, point.data))
                .collect();
            points_to_remove.iter().for_each(|point| {
                rtree.remove(&point);
            });
            geo_faceted_doc_ids -= docids_to_remove;
```

milli/src/update/index_documents/extract/extract_geo_points.rs
```diff
@@ -0,0 +1,45 @@
...
+        create_writer(indexer.chunk_compression_type, indexer.chunk_compression_level, file)
+    })?;
+
+    while let Some((docid_bytes, value)) = obkv_documents.next()? {
+        let obkv = obkv::KvReader::new(value);
+        let point = match obkv.get(geo_field_id) {
+            Some(point) => point,
+            None => continue,
+        };
+        let point: Value = serde_json::from_slice(point).map_err(InternalError::SerdeJson)?;
```
Kerollmops, line 29:
```suggestion
        let point: Value = match obkv.get(geo_field_id) {
            Some(point) => serde_json::from_slice(point).map_err(InternalError::SerdeJson)?,
            None => continue,
        };
```

milli/src/update/index_documents/extract/mod.rs
```diff
@@ -177,6 +185,22 @@ fn extract_documents_data(
...
+    if let Some(geo_field_id) = geo_field_id {
+        let documents_chunk_cloned = documents_chunk.clone();
+        let lmdb_writer_sx_cloned = lmdb_writer_sx.clone();
+        rayon::spawn(move || {
+            let _ = match extract_geo_points(
+                documents_chunk_cloned,
+                indexer,
+                primary_key_id,
+                geo_field_id,
+            ) {
```
Kerollmops, line 197:
```suggestion
            let result = extract_geo_points(documents_chunk_cloned, indexer, primary_key_id, geo_field_id);
            let _ = match result {
```

milli/src/update/index_documents/mod.rs
```diff
@@ -228,11 +228,22 @@ impl<'t, 'u, 'i, 'a> IndexDocuments<'t, 'u, 'i, 'a> {
...
         // get filterable fields for facet databases
         let faceted_fields = self.index.faceted_fields_ids(self.wtxn)?;
+        // get the fid of the `_geo` field.
+        let geo_field_id = if let Some(gfid) = self.index.fields_ids_map(self.wtxn)?.id("_geo") {
+            (self.index.sortable_fields_ids(self.wtxn)?.contains(&gfid)
+                || self.index.filterable_fields_ids(self.wtxn)?.contains(&gfid))
+            .then(|| gfid)
+        } else {
+            None
+        };
```
Kerollmops, line 246:
```suggestion
        let geo_field_id = match self.index.fields_ids_map(self.wtxn)?.id("_geo") {
            Some(gfid) => {
                let is_sortable = self.index.sortable_fields_ids(self.wtxn)?.contains(&gfid);
                let is_filterable = self.index.filterable_fields_ids(self.wtxn)?.contains(&gfid);
                if is_sortable || is_filterable {
                    Some(gfid)
                } else {
                    None
                }
            },
            None => None,
        };
```

milli/src/update/index_documents/typed_chunk.rs
```diff
@@ -177,6 +179,26 @@ pub(crate) fn write_typed_chunk_into_index(
             )?;
             is_merged_database = true;
         }
+        TypedChunk::GeoPoints(mut geo_points) => {
+            // TODO: we should create the rtree with the `RTree::bulk_load` function
```
Kerollmops, line 183: Do we want to keep this TODO here? In which case don't we want to create an issue about that? Do we wait for guigui to try the feature first?

milli/src/update/index_documents/typed_chunk.rs
```diff
@@ -177,6 +179,26 @@ pub(crate) fn write_typed_chunk_into_index(
...
+            let mut rtree = index.geo_rtree(wtxn)?.unwrap_or_default();
+            let mut doc_ids = index.geo_faceted_documents_ids(wtxn)?;
+
+            while let Some((key, value)) = geo_points.next()? {
+                // convert the key back to a u32 (4 bytes)
+                let (key, _) = helpers::try_split_array_at::<u8, 4>(key).unwrap();
+                let key = u32::from_be_bytes(key);
```
Kerollmops, line 190:
```suggestion
                // convert the key back to a u32 (4 bytes)
                let docid = key.try_into().map(DocumentId::from_be_bytes).unwrap();
```

milli/src/update/index_documents/typed_chunk.rs
```diff
@@ -177,6 +179,26 @@ pub(crate) fn write_typed_chunk_into_index(
             )?;
             is_merged_database = true;
         }
+        TypedChunk::GeoPoints(mut geo_points) => {
+            // TODO: we should create the rtree with the `RTree::bulk_load` function
+            let mut rtree = index.geo_rtree(wtxn)?.unwrap_or_default();
+            let mut doc_ids = index.geo_faceted_documents_ids(wtxn)?;
```
Kerollmops, line 185:
```suggestion
            let mut geo_faceted_docids = index.geo_faceted_documents_ids(wtxn)?;
```

Kerollmops (changes requested):
It is a very good PR. It seems to be very well tested and overall nothing more to say than just some styling issues 🍳

milli/src/search/criteria/geo.rs
```diff
@@ -0,0 +1,115 @@
...
+    }
+}
+
+impl<'t> Criterion for Geo<'t> {
+    fn next(&mut self, params: &mut CriterionParameters) -> Result<Option<CriterionResult>> {
+        // if there is no rtree we have nothing to returns
+        let rtree = match self.rtree.as_ref() {
+            Some(rtree) => rtree,
+            None => return Ok(None),
+        };
```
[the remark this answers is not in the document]
irevoire: Oh yes, we chatted about that but I then forgot to do it! My bad

milli/src/search/facet/filter_condition.rs
```diff
@@ -156,6 +161,65 @@ impl FilterCondition {
...
+                        },
+                    // we want to point to the last parameters and if there was no parameters we
+                        // point to the parenthesis
+                    parameters.last().map(|param| param.1.clone()).unwrap_or(param_span),
+                            )))?;
+        }
+        let (lat, lng, distance) = (&parameters[0], &parameters[1], parameters[2].0);
+        if let Some(span) = (!(-181.0..181.).contains(&lat.0))
+            .then(|| &lat.1)
+            .or((!(-181.0..181.).contains(&lng.0)).then(|| &lng.1))
```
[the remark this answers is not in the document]
irevoire: Yes but here I also extract the span indicating if the problem was the latitude or a longitude :eyes:

milli/src/search/facet/filter_condition.rs
irevoire: Currently it's written in the [spec](https://github.com/meilisearch/specifications/blob/develop/text/0059-geo-search.md) currently, ex:
- `_geoRadius({lat}, {lng}, {distance_in_meters})`
- `This field represents the distance in meters of the document from the specified _geoPoint`

milli/src/search/facet/grammar.pest
irevoire: Yes, it'll be more verbose but it should be better in the end! :grin:

milli/src/update/index_documents/typed_chunk.rs
irevoire: I would say we keep the TODO, and we create an issue once it's merged :thinking: 
But as you want, I don't know the rules about letting `TODO`s hanging in the code :grin:

Kerollmops: That's not really something that we do a lot, I would prefer that we just open an issue, please :)

milli/src/search/facet/filter_condition.rs
irevoire: Look like rustfmt just ignore this part of the file :thinking: 
I re-indented it in a more common way :+1:

milli/src/search/criteria/geo.rs
```diff
@@ -0,0 +1,115 @@
...
+
+fn geo_point(
+    rtree: &RTree<GeoPoint>,
+    candidates: RoaringBitmap,
+    point: [f64; 2],
+) -> Box<dyn Iterator<Item = RoaringBitmap>> {
+    let results = rtree
+        .nearest_neighbor_iter(&point)
+        .filter_map(move |point| candidates.contains(point.data).then(|| point.data))
+        .map(|id| std::iter::once(id).collect::<RoaringBitmap>())
```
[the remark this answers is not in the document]
irevoire: done, there was another place in the same file with the same problem

milli/src/update/delete_documents.rs
irevoire: Funny, rust was not understanding this collect but `unzip` worked

milli/src/search/criteria/mod.rs
```diff
@@ -290,18 +292,29 @@ impl<'t> CriteriaBuilder<'t> {
...
                                 )?),
+                                AscDescName::Asc(Member::Geo(point)) => Box::new(Geo::new(
+                                    &self.index,
+                                    &self.rtxn,
+                                    criterion,
+                                    point.clone(),
+                                )?),
+                                AscDescName::Desc(Member::Geo(_point)) => {
+                                    return Err(UserError::InvalidSortName {
+                                        name: "Sorting in descending order is currently not supported for the geosearch".to_string(),
```
Kerollmops, line 315:
```suggestion
                                        name: "sorting in descending order is not supported for the geosearch".to_string(),
```

I am not sure we want to support it in the futur, that's not well defined.

milli/src/search/criteria/geo.rs
```diff
@@ -48,7 +48,14 @@ impl Criterion for Geo<'_> {
...
             Some(rtree) => rtree,
-            None => return Ok(None),
+            None => {
+                return Ok(Some(CriterionResult {
+                    query_tree: None,
+                    candidates: None,
+                    filtered_candidates: Some(params.excluded_candidates.clone()),
+                    bucket_candidates: Some(self.bucket_candidates.clone()),
+                }))
+            }
```
Kerollmops, line 58: Is it the solution chosen with @ManyTheFish?

Kerollmops (changes requested):

milli/src/lib.rs
```diff
@@ -143,7 +143,7 @@ where
 }
 
 /// Return the distance between two points in meters.
-fn distance_between_two_points(a: &[f64; 2], b: &[f64; 2]) -> f64 {
+pub fn distance_between_two_points(a: &[f64; 2], b: &[f64; 2]) -> f64 {
```
Kerollmops, line 146: Can we add a description that specifies that the points are long/lat points, please?

Kerollmops (changes requested):
If that's the solution we chose for the geo sort, that's a great PR so far. Not merging it for now as we found a potential bug with the indexer.

milli/src/search/criteria/geo.rs
irevoire: No sorry I was doing something else after our discussion with @ManyTheFish.
Now the chosen solution is implemented :smile:

milli/src/search/criteria/mod.rs
irevoire: I updated the message

milli/src/lib.rs
irevoire: done

curquiza:
Is this PR breaking?

irevoire:
It adds a bunch of error codes that meilisearch needs to handle but that’s all I think

milli/src/search/criteria/geo.rs
```diff
@@ -0,0 +1,117 @@
...
+fn geo_point(
+    rtree: &RTree<GeoPoint>,
+    candidates: RoaringBitmap,
+    point: [f64; 2],
+) -> Box<dyn Iterator<Item = RoaringBitmap>> {
+    let results = rtree
+        .nearest_neighbor_iter(&point)
+        .filter_map(move |point| candidates.contains(point.data).then(|| point.data))
+        .map(|id| iter::once(id).collect::<RoaringBitmap>())
+        .collect::<Vec<_>>();
```
Kerollmops, line 114: We should stop iterating through the whole list of points, instead, we should stop when all the candidates have been found. By doing a `candidates.remove().is_some()` instead of a simple contains.

Kerollmops (changes requested):

milli/src/search/criteria/geo.rs
irevoire: Done thank you!

milli/src/search/criteria/geo.rs
```diff
@@ -104,12 +104,12 @@ impl Criterion for Geo<'_> {
...
 fn geo_point(
     rtree: &RTree<GeoPoint>,
-    candidates: RoaringBitmap,
+    mut candidates: RoaringBitmap,
     point: [f64; 2],
 ) -> Box<dyn Iterator<Item = RoaringBitmap>> {
     let results = rtree
         .nearest_neighbor_iter(&point)
-        .filter_map(move |point| candidates.contains(point.data).then(|| point.data))
+        .filter_map(move |point| candidates.remove(point.data).then(|| point.data))
```
Kerollmops, line 137: This doesn't make the iteration stop, you should add something like a `take_while` or even better: rewrite this iterator combination as a for loop, instead.

irevoire: oops yes, is this ok?
https://github.com/meilisearch/milli/pull/322/commits/91ce4d1721765cb5862f6cf531fea0d2021021f0

curquiza:
> It adds a bunch of error codes that meilisearch needs to handle but that’s all I think

So it's breaking if I understand well
I add the label :)

milli/src/error.rs
```diff
@@ -12,6 +12,10 @@ use crate::{DocumentId, FieldId};
 
 pub type Object = Map<String, Value>;
 
+pub fn is_reserved_keyword(keyword: &str) -> bool {
+    ["_geo", "_geoDistance"].contains(&keyword)
```
ManyTheFish, line 16: `_geoPoint` seems to leak here; you defined it in the pest grammar.

milli/src/update/index_documents/typed_chunk.rs
```diff
@@ -177,6 +180,24 @@ pub(crate) fn write_typed_chunk_into_index(
             )?;
             is_merged_database = true;
         }
+        TypedChunk::GeoPoints(mut geo_points) => {
+            let mut rtree = index.geo_rtree(wtxn)?.unwrap_or_default();
```
ManyTheFish, line 184: Do we have an idea of the rtree size? I'm a bit worried about smaller VMs, this rtree will grow on each document indexing; and because it's an uncontrolled memory usage, it could make meilisearch be killed.

ManyTheFish (changes requested):

milli/src/update/index_documents/typed_chunk.rs
irevoire: Yes, this will be a problem :pensive: 
Currently, for 1 million documents with a geopoint it looks like the `rtree` consume around 400Mb

ManyTheFish: OUCH! @curquiza we may add a warning in the documentation 😬, some indexing could crash because of it.

irevoire: I'm not sure, but I think @gmourier told the doc team about that and asked to put a big `WARNING: BETA` flag on top of the documentation or something like that? :thinking:

milli/src/error.rs
```diff
@@ -221,7 +227,15 @@ impl fmt::Display for UserError {
...
+                f,
+                "the document with the id: {} contains an invalid _geo field: {}",
+                document_id, object
+            ),
+            Self::InvalidRankingRuleName { name } => write!(f, "invalid criterion {}", name),
+            Self::InvalidReservedRankingRuleName { name } => {
+                write!(f, "{} is a reserved keyword and thus can't be used as a ranking rule", name)
+            }
             Self::InvalidDocumentId { document_id } => {
                 let json = serde_json::to_string(document_id).unwrap();
```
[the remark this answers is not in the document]
irevoire: That's what was done before for the `InvalidCriterionName` so I just kept the same print :grin:

milli/src/error.rs
irevoire: Oops, yes, effectively. And `_geoRadius` was also handled in the pest grammar as a bad usage, so it's pretty logical to put it here too.

Thank you, I added this in this [commit](https://github.com/meilisearch/milli/pull/322/commits/f4b8e5675d903161b262910126cbf6ebf461f89a) with a bunch of tests :smile:

milli/src/criterion.rs
```diff
@@ -50,32 +50,98 @@ impl FromStr for Criterion {
...
+
+impl FromStr for Member {
+    type Err = UserError;
+
+    fn from_str(text: &str) -> Result<Member, Self::Err> {
+        if text.starts_with("_geoPoint(") {
+            let point =
+                text.strip_prefix("_geoPoint(")
+                    .and_then(|point| point.strip_suffix(")"))
+                    .ok_or_else(|| UserError::InvalidRankingRuleName { name: text.to_string() })?;
```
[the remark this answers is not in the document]
irevoire: Ah yes thank you!
I did it in this [commit](https://github.com/meilisearch/milli/pull/322/commits/f4b8e5675d903161b262910126cbf6ebf461f89a) like that:
```rust
        if let Some(point) = text.strip_prefix("_geoPoint(").and_then(|text| text.strip_suffix(")"))
        {
```

irevoire:
When writing unit tests for the `AscDesc::from_str` function I noticed how we were accepting strange syntax.
This is not the PR to fix everything but maybe @gmourier should look at that and we should only accept a certain subset of characters instead of allowing anything.
All of the following strings are considered as valid sort rules or criterion currently:
```
"bidule:desc"
"a-b:desc"
"a:b:desc"
"a12:asc"
"42:asc"
"_geoPoint(42, 59):asc"
"_geoPoint(42.459, 59):desc"
"_geoPoint(42, 59.895):desc"
"_geoPoint(42, 59.895):desc"
"_geoPoint(42.0002, 59.895):desc"
"_geoPoint(42., 59.):desc"
"truc(12, 13):desc"
```

milli/src/criterion.rs
```diff
@@ -50,32 +50,95 @@ impl FromStr for Criterion {
...
+                        .map_err(|_| UserError::InvalidRankingRuleName { name: text.to_string() })
+                })?;
+            Ok(Member::Geo([lat, long]))
+        } else {
+            if is_reserved_keyword(text) {
+                return Err(UserError::InvalidReservedRankingRuleName { name: text.to_string() })?;
+            }
+            Ok(Member::Field(text.to_string()))
+        }
+    }
```
Kerollmops, line 97: I would prefer that you use a match instead ;)

```suggestion
        match text.strip_prefix("_geoPoint(").and_then(|text| text.strip_suffix(")")) {
            Some(point) => {
                let (lat, long) = point
                  .split_once(',')
                  .ok_or_else(|| UserError::InvalidRankingRuleName { name: text.to_string() })
                  .and_then(|(lat, long)| {
                      lat.trim()
                          .parse()
                          .and_then(|lat| long.trim().parse().map(|long| (lat, long)))
                          .map_err(|_| UserError::InvalidRankingRuleName { name: text.to_string() })
                  })?;
              Ok(Member::Geo([lat, long]))
        },
        None => {
            if is_reserved_keyword(text) {
                return Err(UserError::InvalidReservedRankingRuleName { name: text.to_string() })?;
            }
            Ok(Member::Field(text.to_string()))
        }
    }
```

Kerollmops (changes requested):
Apart from this little change, I think this is ready to be merged.

Kerollmops (approved):
That's ready to merge! Thank you, Tamo!

Kerollmops:
bors merge

ManyTheFish (approved):

ManyTheFish:
bors merge

PR #324, opened 2021-08-24
implement documents api

MarinPostma:


milli/src/documents/serde.rs
```diff
@@ -0,0 +1,465 @@
...
+use std::{fmt, io};
+
+use byteorder::{BigEndian, WriteBytesExt};
+use obkv::KvWriter;
+use serde::ser::{Impossible, Serialize, SerializeMap, SerializeSeq, Serializer};
+
+use super::{AdditionIndex, ByteCounter, Error};
+use crate::FieldId;
+
+pub struct DocumentsSerilializer<W> {
```
Kerollmops, line 11:
```suggestion
pub struct DocumentsSerializer<W> {
```

milli/src/heed_codec/facet/facet_level_value_f64_codec.rs
```diff
@@ -34,11 +34,11 @@ impl heed::BytesEncode<'_> for FacetLevelValueF64Codec {
     fn bytes_encode((field_id, level, left, right): &Self::EItem) -> Option<Cow<[u8]>> {
         let mut buffer = [0u8; 32];
 
-        let len = if *level != 0 {
```
Kerollmops, line 37: Was it necessary to change that in this PR? Can you rollback this, please?

milli/src/heed_codec/facet/facet_string_level_zero_value_codec.rs
```diff
@@ -34,7 +34,7 @@ where
     type EItem = (&'a str, C::EItem);
 
     fn bytes_encode((string, value): &'a Self::EItem) -> Option<Cow<[u8]>> {
-        let value_bytes = C::bytes_encode(&value)?;
```
Kerollmops, line 37: Same here, can you rollback, please?

milli/src/heed_codec/facet/facet_string_zero_bounds_value_codec.rs
```diff
@@ -66,14 +66,14 @@ where
                 bytes.extend_from_slice(left.as_bytes());
                 bytes.extend_from_slice(right.as_bytes());
 
-                let value_bytes = C::bytes_encode(&value)?;
```
Kerollmops, line 69: can you roolback, please?

milli/src/heed_codec/facet/facet_level_value_f64_codec.rs
Kerollmops: Is it due to a bump of clippy?

milli/src/update/index_documents/mod.rs
```diff
@@ -136,16 +124,17 @@ impl<'t, 'u, 'i, 'a> IndexDocuments<'t, 'u, 'i, 'a> {
...
-        R: io::Read,
+        R: Read + Seek,
         F: Fn(UpdateIndexingStep, u64) + Sync,
     {
-        let mut reader = BufReader::new(reader);
-        reader.fill_buf()?;
-
         // Early return when there is no document to add
-        if reader.buffer().is_empty() {
+        if reader.len() == 0 {
```
Kerollmops, line 137: Can you add a `is_empty` method and use it here, please? Clippy should have warned you about that :)

milli/src/documents/builder.rs
```diff
@@ -0,0 +1,64 @@
...
+        bincode::serialize_into(&mut writer, &meta)?;
+
+        writer.seek(io::SeekFrom::Start(0))?;
+        writer.write_u64::<BigEndian>(offset as u64)?;
+
+        writer.flush()?;
+
+        Ok(())
+    }
+
```
Kerollmops, line 53: I would like that you introduce [an `into_inner` method](https://docs.rs/fst/0.4.7/fst/struct.MapBuilder.html#method.into_inner) that consumes `self` and call `finish` before returning the inner writer. It is a better way to use this kind of wrapper type, the user gives the writer to the `DocumentsBuilder` and when you are done, you ask the wrapper type to give you back your writer, this way you can't forget to call finish.

milli/src/documents/mod.rs
```diff
@@ -0,0 +1,231 @@
...
+impl std::error::Error for Error {}
+
+/// Macro used to generate documents, with the same syntax as `serde_json::json`
+#[cfg(test)]
+macro_rules! documents {
+    ($data:tt) => {{
+        let documents = serde_json::json!($data);
+        let mut writer = std::io::Cursor::new(Vec::new());
+        let mut builder =
+            crate::documents::DocumentsBuilder::new(&mut writer, bimap::BiHashMap::new()).unwrap();
```
Kerollmops: For example, here we should not use `&mut writer` but rather `writer` directly and call `into_inner` to get the `writer` back without calling `finish` on it.

milli/src/update/index_documents/transform.rs
```diff
@@ -51,90 +53,63 @@ pub struct Transform<'t, 'i> {
...
-        reader: R,
-        progress_callback: F,
-    ) -> Result<TransformOutput>
-    where
-        R: Read,
-        F: Fn(UpdateIndexingStep) + Sync,
-    {
-        self.output_from_generic_json(reader, true, progress_callback)
-    }
+fn find_primary_key(index: &bimap::BiHashMap<u16, String>) -> Option<&str> {
```
Kerollmops, line 79: What's that `u16`? The `FieldId` type?

milli/src/update/index_documents/transform.rs
```diff
@@ -151,204 +126,99 @@ impl Transform<'_, '_> {
...
-                progress_callback(UpdateIndexingStep::TransformFromUserIntoGenericFormat {
+                progress_callback(UpdateIndexingStep::RemapDocumentAddition {
                     documents_seen: documents_count,
                 });
             }
 
-            obkv_buffer.clear();
-            let mut writer = obkv::KvWriter::<_, FieldId>::new(&mut obkv_buffer);
+            let mut external_id_buffer = SmallVec::<[u8; 512]>::new();
+            let mut field_buffer = SmallVec::<[(u16, &[u8]); 128]>::new();
```
Kerollmops, line 139: I would like to get rid of these `SmallVec` things and use a normal `Vec` here. Could you also remove the `write` feature on the samllvec in the Cargo.toml too?

milli/src/update/index_documents/transform.rs
```diff
@@ -151,204 +126,99 @@ impl Transform<'_, '_> {
...
-                    Value::Number(number) => Cow::Owned(number.to_string()),
-                    content => {
-                        return Err(
-                            UserError::InvalidDocumentId { document_id: content.clone() }.into()
-                        )
-                    }
-                },
+            // We need to make sure that every document has a primary key. After we have remapped
+            // all the fields in the document, we try to find the primary key value. If we can find
+            // it, transform it into a string and validate it, and then update it in the the
```
Kerollmops, line 149:
```suggestion
            // it, transform it into a string and validate it, and then update it in the
```

milli/src/update/index_documents/transform.rs
```diff
@@ -151,204 +126,99 @@ impl Transform<'_, '_> {
...
+            field_buffer.sort_unstable_by(|(f1, _), (f2, _)| f1.cmp(&f2));
 
-            // We retrieve the field id based on the fields ids map fields ids order.
-            for (field_id, field) in iter {
-                // We serialize the attribute values as JSON strings.
-                json_buffer.clear();
-                serde_json::to_writer(&mut json_buffer, &field)
-                    .map_err(InternalError::SerdeJson)?;
-                writer.insert(*field_id, &json_buffer)?;
+            // The last step is to build the new new obkv document, and insert it in the sorter.
```
Kerollmops, line 204:
```suggestion
            // The last step is to build the new obkv document, and insert it in the sorter.
```

milli/src/update/index_documents/transform.rs
```diff
@@ -151,204 +126,99 @@ impl Transform<'_, '_> {
...
+            let mut writer = obkv::KvWriter::new(&mut obkv_buffer);
+            for (k, v) in field_buffer.iter() {
+                writer.insert(*k, v)?;
             }
 
             // We use the extracted/generated user id as the key for this document.
-            sorter.insert(external_id, &obkv_buffer)?;
+            sorter.insert(&external_id.as_ref().as_bytes(), &obkv_buffer)?;
```
Kerollmops, line 211:
```suggestion
            sorter.insert(&external_id, &obkv_buffer)?;
```

Not sure why you need to do all of this, but maybe just doing this is OK as the insert method accepts an `AsRef<[u8]>`.

milli/src/update/settings.rs
```diff
@@ -695,37 +711,43 @@ mod tests {
...
+        for document in index.all_documents(&rtxn).unwrap() {
+            let document = document.unwrap();
+            let json = crate::obkv_to_json(&fidmap.ids().collect::<Vec<_>>(), &fidmap, document.1)
+                .unwrap();
+            println!("json: {:?}", json);
+        }
         let count = index
             .facet_id_f64_docids
             .remap_key_type::<ByteSlice>()
             // The faceted field id is 2u16
```
Kerollmops, line 725:
```suggestion
            // The faceted field id is 1u16
```

milli/src/update/update_step.rs
```diff
@@ -2,10 +2,9 @@ use UpdateIndexingStep::*;
 
 #[derive(Debug, Clone, Copy)]
 pub enum UpdateIndexingStep {
-    /// Transform from the original user given format (CSV, JSON, JSON lines)
-    /// into a generic format based on the obkv and grenad crates. This step also
-    /// deduplicate potential documents in this batch update by merging or replacing them.
-    TransformFromUserIntoGenericFormat { documents_seen: usize },
+    /// Remap document addition fields the one present in the database, adding new fields in to the
+    /// schema on the go.
```
Kerollmops, line 6: Could you rephrase this, please?

milli/tests/search/mod.rs
```diff
@@ -53,9 +56,21 @@ pub fn setup_search_index_with_criteria(criteria: &[Criterion]) -> Index {
...
+    let mut cursor = Cursor::new(Vec::new());
+    let mut documents_builder =
+        DocumentsBuilder::new(&mut cursor, bimap::BiHashMap::new()).unwrap();
+    let reader = Cursor::new(CONTENT.as_bytes());
+    for doc in serde_json::Deserializer::from_reader(reader).into_iter::<serde_json::Value>() {
+        documents_builder.add_documents(doc.unwrap()).unwrap();
+    }
+    documents_builder.finish().unwrap();
+
+    cursor.set_position(0);
```
Kerollmops: I am not sure you need to wrap the vec into a cursor [as `&mut Vec<u8>` implements write](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Write). And [you can read it by using a `&[u8]`](https://doc.rust-lang.org/std/primitive.slice.html#impl-Read), but that's maybe easier to just wrap it into a `cursor`.

Kerollmops (changes requested):
I didn't review your clippy commit as it is too much and must not be in this PR, can you please remove it from there and create a new PR by cherry-picking it, please?

milli/tests/search/mod.rs
MarinPostma: It doesn't implement Seek

milli/src/documents/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+/// provides an API to easily build and read such documents.
+///
+/// The `DocumentBuilder` interface allows to write batches of documents to a writer, that can
+/// later be read by milli using the `DocumentsReader` interface.
+mod reader;
+mod serde;
+
+use std::{fmt, io};
+
+use ::serde::{Deserialize, Serialize};
```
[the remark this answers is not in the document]
MarinPostma: I have a module called serde, this needed

milli/src/documents/mod.rs
```diff
@@ -0,0 +1,233 @@
+mod builder;
+/// The documents module defines an intermediary document format that milli uses for indexation, and
+/// provides an API to easily build and read such documents.
+///
+/// The `DocumentBuilder` interface allows to write batches of documents to a writer, that can
```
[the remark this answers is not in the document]
MarinPostma: The 's' is purposedly here, it represents a set of documents, not a single one.

milli/src/documents/mod.rs
MarinPostma: yes but nothing prevents you from doing it, so it's not a strong guarantee...

milli/src/update/index_documents/transform.rs
MarinPostma: So you would rather allocate all the time? what's the rationale?

milli/src/documents/builder.rs
```diff
@@ -0,0 +1,63 @@
...
+    /// Returns the number of documents that have been written to the builder.
+    pub fn len(&self) -> usize {
+        self.serializer.count
+    }
+
+    /// This method must be called after the document addition is terminated. It will put the
+    /// metadata at the end of the file, and write the metadata offset at the beginning on the
+    /// file.
+    pub fn finish(self) -> Result<(), Error> {
```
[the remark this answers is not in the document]
MarinPostma: I don't believe  this is a good idea, you mentioned the fst crate and into_inner, and they don't implement drop either. Ignoring an error happening here means we you wrote a corrupted file and don't know about it.

http-ui/src/main.rs
```diff
@@ -835,7 +835,7 @@ async fn main() -> anyhow::Result<()> {
             UpdateStatus<UpdateMeta, UpdateMetaProgress, String>,
         >,
         update_method: Option<String>,
-        update_format: UpdateFormat,
+        format: String,
```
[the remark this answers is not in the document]
MarinPostma: this is more complicated than that, i need to pass this string around until I put it into a UpdateMeta. Having a 'static lifetime in update meta causes conflicts with the lifetime requirement for the update store that require `for<'a> Deserialize<'a>` as a bound. It is not worth the trouble tryng to find a fix for it for http-ui...

milli/src/documents/builder.rs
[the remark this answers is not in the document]
Kerollmops: The downside of your solution is that it is also possible to just forget to use the `finish` method and that would be worse than panicking in the `Drop` trait impl.

milli/src/documents/mod.rs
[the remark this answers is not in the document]
Kerollmops: Won't `self::serde` work instead?

milli/src/documents/mod.rs
Kerollmops: If you don't call `into_inner` here you will not be able to retrieve your `writer` and write it in the obkv, so yeah it prevents you from forgetting to call `into_inner` (where we flush on disk).

milli/src/update/index_documents/transform.rs
Kerollmops: Allocating is OK here, that's a lot of boilerplate, maybe we can just reuse the `Vec<u8>` buffer by `clear`ing it like before. Adding smallvec just for this is too much boilerplate.

milli/src/documents/builder.rs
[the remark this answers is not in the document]
MarinPostma: That also means that you would call Finnish even if you decide to drop the an unfinished addition, for example if an error occured while adding documents, which doesn't make much sense

I was thinking that we could check is the reader was finished with an assertion in the drop, but then you can't drop an unfinished builder, which is not acceptable either

milli/src/documents/mod.rs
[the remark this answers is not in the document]
MarinPostma: Self::serde refers to the serde in my module, not serde

milli/src/update/index_documents/transform.rs
MarinPostma: No, this is why I used smallvec here, rust thinks that I'm keeping references in my vecs across loop iterations

milli/src/update/index_documents/transform.rs
MarinPostma: Cow<&str> doesn't implement AsRef<u8> unfortunately

milli/src/documents/builder.rs
MarinPostma: Well, in real usecases you don't need into_inner, you will be writing the documents to a file, and later on, re-open the file and read from it, so for real use cases, it would be even worse.

The only place where it could make sense is for tests, then then again, if you forget to call finish, your test will fail...

milli/src/update/index_documents/transform.rs
Kerollmops: So, if you really want to keep these `SmallVec`s, please declare type aliases like `SmallVec512<T>` in the lib.rs. But I maintain that it is not a good thing to put 512 + (2 + 8) * 128 = 1 792 bytes on the stack.

Kerollmops: I would prefer that you declare a Vec buffer above the loop (131L) and clear it before we use it, like for the `obkv_buffer` before.

Kerollmops (changes requested):

milli/src/update/index_documents/transform.rs
Kerollmops: `Cow<str>` does implement `AsRef<str>` which it self implements `Cow<[u8]>`, but is [it better](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bcd13d7ac7e6ddf4d1c1a5c46a515ba1)?

milli/src/update/index_documents/transform.rs
MarinPostma: You are right for the `external_id_buffer`. The issue with the `field_buffer` is that it holds references. I am not super happy with this solution either. I have made some ressearches:
- The first solution is to use unsafe code, and transmute the Vec to make the lifetimes match (after making sure that the vector is empty)
- We can use this crate: https://docs.rs/vec-utils/0.3.0/vec_utils/trait.VecExt.html#tymethod.drop_and_reuse to do it safely.

MarinPostma: @Kerollmops I made a new commit with the proposed solution, tell me what you think :)

Kerollmops: I must admit that I largely prefer this solution, even if importing a new dependency makes me feel bad :feelsgood:, at least it doesn't have any other dependency itself.

The problem is due to the `(u16, &[u8])` type keeping the lifetime metadata, I tried declaring an `fn move_me<T>(vec) -> Vec<T>` function but it wasn't erasing the lifetime meta as the `T` type wasn't altered, [the `drop_and_reuse` method](https://docs.rs/vec-utils/0.3.0/vec_utils/trait.VecExt.html#tymethod.drop_and_reuse) is changing from a `T` to a `U` type, that helps the compiler erase the lifetime.

MarinPostma: @Kerollmops, according to https://users.rust-lang.org/t/pattern-how-to-reuse-a-vec-str-across-loop-iterations/61657/4, you can do that:

```rust
fn reuse_vec<T, U>(mut v: Vec<T>) -> Vec<U> {
    assert_eq!(std::mem::size_of::<T>(), std::mem::size_of::<U>());
    assert_eq!(std::mem::align_of::<T>(), std::mem::align_of::<U>());
    v.clear();
    v.into_iter().map(|_| unreachable!()).collect()
}
```

which only works because of the optimization done in this pr https://github.com/rust-lang/rust/pull/70793

now if that's not black magic? haha

We can just use this function, but since it's rather confusing why it works, maybe we can keep the dependency. It's up to you.

Kerollmops: That's interesting but, as you say, too much black magic. We can keep this crate but please move it in the right place in the Cargo.toml dependencies 😃 What a story, I tried for maybe 30mins to make it work without `unsafe` and any external dependencies!

ManyTheFish (changes requested):

ManyTheFish (changes requested):

milli/src/documents/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+    fn from(other: io::Error) -> Self {
+        Self::Io(other)
+    }
+}
+
+impl From<bincode::Error> for Error {
+    fn from(other: bincode::Error) -> Self {
+        Self::Serialize(other)
+    }
+}
```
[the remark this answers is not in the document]
MarinPostma: I thought about it for a moment, but the document format resides a bit in between milli and a client, they may each have to handle errors in their own way, so I separated to error from the actual milli error. I can put the in a submodule error if you prefer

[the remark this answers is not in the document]
ManyTheFish: It's ok like this

ManyTheFish (approved):

MarinPostma:
I have rebased and squashed this pr

milli/Cargo.toml
```diff
@@ -33,7 +34,7 @@ serde = { version = "1.0.123", features = ["derive"] }
 serde_json = { version = "1.0.62", features = ["preserve_order"] }
 slice-group-by = "0.2.6"
 smallstr =  { version = "0.2.0", features = ["serde"] }
-smallvec = "1.6.1"
+smallvec = { version = "1.6.1", features = ["write"] }
```
Kerollmops, line 37: Do we need this? I think we can remove the `write` feature now.

milli/src/update/index_documents/transform.rs
```diff
@@ -610,6 +482,17 @@ fn validate_document_id(document_id: &str) -> Option<&str> {
...
+/// Drops all the value of type `U` in vec, and reuses the allocation to create a `Vec<T>`.
+///
+/// The size and alignment of T and U must match.
+fn drop_and_reuse<U, T>(mut vec: Vec<U>) -> Vec<T> {
+    debug_assert_eq!(std::mem::align_of::<U>(), std::mem::align_of::<T>());
+    debug_assert_eq!(std::mem::size_of::<U>(), std::mem::size_of::<T>());
+    vec.clear();
+    debug_assert!(vec.is_empty());
+    vec.into_iter().map(|_| unreachable!()).collect()
+}
```
Kerollmops, line 494: Ok, that's what I thought would be good to do instead of importing the whole library and you did it, thank you!

Kerollmops (changes requested):

milli/Cargo.toml
MarinPostma: done

Kerollmops (approved):
Thank you marin!
bors merge

PR #335, opened 2021-08-31
Get sortable_fields from index only if criteria present in query

shekhirin:
Seems like we don't need to retrieve `sortable_fields` from the index if there's no any `sort_criteria` in the query.

Small 🤏  optimization opportunity out there.


Kerollmops (approved):
Thank you! LGTM 🚀 
bors merge

PR #341, opened 2021-09-07
Throw a query time error when a sort parameter is used but the sort ranking rule is missing

Kerollmops:
This PR makes the engine throw an error for when the ranking rules don't contain the `sort` rule, the `sortable_fields` are correctly set but the user tries to use the `sort` query parameter. Doing so will have no effect on the returned documents so we preferred returning an error to help debug this.

That's breaking on the MeiliSearch side as we added a new variant to the `UserError` enum.


irevoire (approved):
code wise, this looks good to me

Kerollmops:
bors merge

Issue #339, opened 2021-09-06
Sending an attribute with a malformed syntax in `sort` at search time does not raise an invalid_sort error


gmourier:
With sortable-attributes

```json
["title", "rates"]
```

If I send `sort` in this way at search time (missing :asc/:desc)
```json
{
    "sort": ["title"]
}
```
or if I send `sort` in this way (malformed :asc/:desc)
```json
{
    "sort": ["title:deesc"]
}
```

The API response is 
```json
{
    "message": "invalid criterion title:deesc",
    "errorCode": "internal",
    "errorType": "internal_error",
    "errorLink": "https://docs.meilisearch.com/errors#internal"
}
```

As a user, I would like to receive an `invalid_sort` error indicating a syntax error in the `sort` parameters.

curquiza:
@irevoire. The patch of the spec describes the message we need :) 
https://github.com/meilisearch/specifications/pull/68#pullrequestreview-747136355

PR #342, opened 2021-09-07
Let the caller decide what kind of error they want to returns when parsing `AscDesc`

irevoire:
This is one possible fix for #339 
We would then need to patch these lines https://github.com/meilisearch/MeiliSearch/blob/main/meilisearch-http/src/index/search.rs#L110-L114 to return the error we want.

Another solution would be to add a parameter to the `from_str` to specify which context we are in.

### DIFF
--- a/milli/src/error.rs
+++ b/milli/src/error.rs
@@ -53,11 +53,13 @@ pub enum UserError {
     AttributeLimitReached,
     Csv(csv::Error),
     DocumentLimitReached,
+    InvalidAscDescSyntax { name: String },
     InvalidCriterionName { name: String },
     InvalidDocumentId { document_id: Value },
     InvalidFacetsDistribution { invalid_facets_name: HashSet<String> },
     InvalidFilter(pest::error::Error<ParserRule>),
     InvalidFilterAttribute(pest::error::Error<ParserRule>),
+    InvalidSortName { name: String },
     InvalidSortableAttribute { field: String, valid_fields: HashSet<String> },
     SortRankingRuleMissing,
     InvalidStoreFile,
@@ -216,6 +218,9 @@ impl fmt::Display for UserError {
                 )
             }
             Self::InvalidFilter(error) => error.fmt(f),
+            Self::InvalidAscDescSyntax { name } => {
+                write!(f, "invalid asc/desc syntax for {}", name)
+            }
             Self::InvalidCriterionName { name } => write!(f, "invalid criterion {}", name),
             Self::InvalidDocumentId { document_id } => {
                 let json = serde_json::to_string(document_id).unwrap();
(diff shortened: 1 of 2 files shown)


Kerollmops (changes requested):
I am not sure it is the best solution, couldn't we introduce a new struct unit error instead, just for the `AscDesc` syntax? i.e. `struct InvalidAscDescSyntax`. We can then just do a `text.to_string()` on the caller side?

irevoire:
How is a `to_string` going to solve our issue?
Do you mean something like a `to_criterion_error` and `to_sort_error`?

Kerollmops:
No, What I want is to create a custom error specific to this error instead of returning a `String` in the `Err` variant. Then when an error is found i.e. Err(`InvalidAscDescSyntax`) we can simply create the error message like before i.e `UserError::InvalidCriterionName` or any other error. I don't like the fact that we return a `text.to_string()` in the error variant, it doesn't mean anything and is useless as we already have the `text` on the caller side.

irevoire:
Ah ok I get the idea thanks!
I'll do it tomorrow morning

milli/src/error.rs
```diff
@@ -216,6 +218,7 @@ impl fmt::Display for UserError {
                 )
             }
             Self::InvalidFilter(error) => error.fmt(f),
+            Self::InvalidAscDescSyntax { name } => write!(f, "invalid asc/desc syntax {}", name),
```
Kerollmops, line 221: Not sure about this but the error message is not the one we want.
https://github.com/meilisearch/specifications/pull/68/files#diff-a3090c4f9494d7b6ef78610abf435f2ce4fcd50b118ec82f81701fbef0be773aR203

milli/src/error.rs
```diff
@@ -228,6 +231,7 @@ only composed of alphanumeric characters (a-z A-Z 0-9), hyphens (-) and undersco
                 )
             }
             Self::InvalidFilterAttribute(error) => error.fmt(f),
+            Self::InvalidSortName { name } => write!(f, "invalid sort {}", name),
```
Kerollmops, line 234: Or even there the message doesn't seem to be the one we want neither.

Kerollmops (changes requested):

milli/src/error.rs
irevoire: You're right!

milli/src/error.rs
```diff
@@ -216,6 +218,7 @@ impl fmt::Display for UserError {
                 )
             }
             Self::InvalidFilter(error) => error.fmt(f),
+            Self::InvalidAscDescSyntax { name } => write!(f, "{}", name),
```
Kerollmops, line 221: Not the expected error message.

irevoire: What message do you want to return?
To me we are not supposed to return anything here so the caller can then do a `to_string` on the error and wrap it into either a InvalidSort or a InvalideCriterion

Kerollmops: Hum, as it is an error I would like to have a message in front of the invalid syntax, something like "invalid asc/desc syntax for {}". The receiver of this error variant can just pattern match and retrieve the `name` field and do what he want.

Kerollmops (approved):
Thank you very much, merging! 🧛
bors merge

milli/src/error.rs
irevoire: done

PR #344, opened 2021-09-07
Move the sort ranking rule before the exactness ranking rule

Kerollmops:
This PR moves the sort ranking rule at the 5th position by default, right before the exactness one.


ManyTheFish (approved):
bors merge

PR #345, opened 2021-09-07
Better dependencies cache for CI

shekhirin:


.github/workflows/rust.yml
```diff
@@ -23,13 +23,7 @@ jobs:
...
     - uses: actions/checkout@v2
     - name: Cache dependencies
-      uses: actions/cache@v2
-      with:
-        path: |
-          ~/.cargo
-          ./Cargo.lock
-          ./target
-        key: ${{ matrix.os }}-${{ matrix.rust }}-${{ hashFiles('Cargo.toml') }}
+      uses: Swatinem/rust-cache@c5ed9ba6b7e1bb8aff90d43acd2f0af4990fa57c
```
curquiza, line 26:
```suggestion
      uses: Swatinem/rust-cache@v1
```

It is possible to use `@v1` instead? :)

shekhirin: I'd not go for it 😄 

> I use the v1 branch similar to master development, so if you want to have a more stable experience, please use a fixed revision or tag.

(https://github.com/Swatinem/rust-cache#versioning)

curquiza:
Hello @shekhirin! Thanks for your PR 🙂

What is the difference with the previous cache system we used in our CIs?

.github/workflows/rust.yml
curquiza: Can we use a version then, I mean a more human readable version, like v1.3.0? :)

shekhirin:
Hey @curquiza,

It caches almost the same directories (https://github.com/Swatinem/rust-cache/blob/c5ed9ba6b7e1bb8aff90d43acd2f0af4990fa57c/src/common.ts#L76-L88), main difference is the clean up in `target` directory and some of `~/.cargo` subdirectories before caching (https://github.com/Swatinem/rust-cache/blob/v1/src/save.ts#L35-L49).

.github/workflows/rust.yml
shekhirin: yep, done!

curquiza:
Thanks for your explanation @shekhirin! :)

I put @irevoire as a final reviewer, as he was the first one integrating the cache to the CI :)

irevoire (approved):
It looks good to me!
It's faster and way more concise, once again thanks @shekhirin :tada: 

bors merge

PR #346, opened 2021-09-07
remove unused grenad default features

MarinPostma:
Milli is not using any of grenad default features, and it's zstd feature creates conflict with meilisearch. This pr simply remove the unused features.


Kerollmops (approved):
I was certain that we were using the snappy compression but doesn't seem to break anything so, LGTM!
bors merge

PR #349, opened 2021-09-08
Enable the grenad tempfile feature back

Kerollmops:
This PR enables the grenad `tempfile` feature back, [when this is feature is disabled the sorter writes the entries in memory](https://github.com/Kerollmops/grenad/blob/7c082d05bf3fe88c4e66a8104e4f01229fb12c67/src/sorter.rs#L470-L476) instead of on disk and therefore, consumes more memory. By enabling this feature grenad merges on disk by using the `tempfile` dependency.

This PR also bumps milli to v0.3.1 where @ManyTheFish added an assert for when the allocator can't allocate and disable the default snappy compression in the `http-ui` crate.


ManyTheFish (approved):
bors merge

PR #359, opened 2021-09-16
Improve the benchmark comparison script

irevoire:
This modification allow us to compare more than 2 benchmarks or to only print the results of one benchmark

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -54,7 +54,7 @@ MILLI_BENCH_DATASETS_PATH=~/datasets cargo bench --bench songs # the code is com
 
 ## Comparison between benchmarks
 
-The benchmark reports we push are generated with `critcmp`. Thus, we use `critcmp` to generate comparison results between 2 benchmarks.
+The benchmark reports we push are generated with `critcmp`. Thus, we use `critcmp` to show the result of a benchmark, or compare results between multiple benchmarks.
 
 We provide a script to download and display the comparison report.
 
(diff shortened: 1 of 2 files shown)


curquiza (commented):
Should we change the README section? https://github.com/meilisearch/milli/tree/main/benchmarks#comparison-between-benchmarks

```
The benchmark reports we push are generated with critcmp. Thus, we use critcmp to generate comparison results between 2 benchmarks.
```
-> still 2?  Seems like more are possible

irevoire:
@curquiza, yes, you're right! I updated the README.
I was also looking for a new name for the script but found nothing really good :unamused:

benchmarks/README.md
```diff
@@ -54,7 +54,7 @@ MILLI_BENCH_DATASETS_PATH=~/datasets cargo bench --bench songs # the code is com
 
 ## Comparison between benchmarks
 
-The benchmark reports we push are generated with `critcmp`. Thus, we use `critcmp` to generate comparison results between 2 benchmarks.
+The benchmark reports we push are generated with `critcmp`. Thus, we use `critcmp` to show the result of a benchmarks, or compare results between multiple benchmarks.
```
curquiza, line 57:
```suggestion
The benchmark reports we push are generated with `critcmp`. Thus, we use `critcmp` to show the result of a benchmark, or compare results between multiple benchmarks.
```

curquiza (dismissed):
Really clear :)

curquiza (approved):
Top moumoute

irevoire:
bors merge

PR #363, opened 2021-09-22
Fix the returned `AscDesc` error

irevoire:
With my previous PR on the geosearch I erased the change I've introduced with my pre-previous PR about the new error type when we fail to parse the `AscDesc` type.

Sorry for that, here is the fix


Kerollmops (approved):
Whooops, thank you Tamo!
bors merge

PR #364, opened 2021-09-22
Fix all the benchmarks 

irevoire:
#324 broke all benchmarks.
I fixed everything and noticed that `cargo check --all` was insufficient to check the bench in multiple workspaces, so I also updated the CI to use `cargo check --workspace --all-targets`.


Kerollmops (approved):
Thank you!
bors merge

PR #369, opened 2021-09-23
Add test checking the bug reported in meilisearch issue 1716

ManyTheFish:
The bug is not present in the newer milli version.

Related to [Meilisearch#1716](https://github.com/meilisearch/MeiliSearch/issues/1716)


Kerollmops (approved):
bors merge

PR #371, opened 2021-09-28
Provide a sort error handler

irevoire:
This PR simplify the error handling of asc-desc rules for Meilisearch or any other wrapper by providing directly in milli a new error type called `SortError` that can be generated from an `AscDescError` and that can be automatically converted to a `UserError`.

Basically now, wherever you are in the code as a user or in milli you can parse an `AscDesc` syntax and depending on the context, cast it either as a `SortError` or a `CriterionError` in one line with improved error messages.


Kerollmops (approved):
That, indeed, helps a lot and makes the code clearer.
Thank you!

bors merge

PR #372, opened 2021-09-28
Fix Meilisearch 1714

ManyTheFish:
The bug comes from the typo tolerance, to know how many typos are accepted we were counting bytes instead of characters in a word.
On Chinese Script characters, we were allowing  2 typos on 3 characters words.
We are now counting the number of char instead of counting bytes to assign the typo tolerance.

Related to [Meilisearch#1714](https://github.com/meilisearch/MeiliSearch/issues/1714)


Kerollmops (approved):
Thank you very much, manyllmops!
bors merge

PR #373, opened 2021-09-28
Improve error message for bad sort syntax with geosearch

irevoire:
@Kerollmops This should be the last PR for the geosearch and error handling, sorry for doing it in so many steps :grimacing:


Kerollmops (approved):
No worries Tamo, ready to be merged!
Nothing to say about that apart from the fact that it seems to introduce breaking changes, can you add the breaking labels to this PR and the previous one if necessary?

bors merge

PR #376, opened 2021-09-28
Stop casting integer docids to string

irevoire:
When a docid is an integer, we stop casting it to a string, and thus we don't add `"` around it.

### DIFF
--- a/milli/src/update/index_documents/mod.rs
+++ b/milli/src/update/index_documents/mod.rs
@@ -536,7 +536,7 @@ mod tests {
 
         // Check that this document is equal to the last one sent.
         let mut doc_iter = doc.iter();
-        assert_eq!(doc_iter.next(), Some((0, &br#""1""#[..])));
+        assert_eq!(doc_iter.next(), Some((0, &b"1"[..])));
         assert_eq!(doc_iter.next(), Some((1, &br#""benoit""#[..])));
         assert_eq!(doc_iter.next(), None);
         drop(rtxn);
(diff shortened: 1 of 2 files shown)


MarinPostma (approved):
Yes! thank you!!!

Kerollmops (commented):
I think I understand the change, we were altering the user documents by converting the document's external ids into strings. With this PR we no more alter the user's document but I am not sure if it works great in the case that two documents have the same document ids but one in a string (`"12345"`) and the other as a rough integer (`12345`)? Does the get document route works great in this case?

MarinPostma:
@Kerollmops This was not the case before, It was an error I have introduced when transcribing the transform functions.

As for your concern, it works as expected :)
image

Kerollmops:
@MarinPostma Ho, ok! So this fixes a recent bug, thank you. However, I was talking about [the get one document API](https://docs.meilisearch.com/reference/api/documents.html#get-one-document) where you need to specify the document id in the URL.

MarinPostma:
@Kerollmops yup
image

Kerollmops (approved):
Thank you Tamo and Marin for the changes! As it works great, I merge it 🚀
bors merge

ManyTheFish:
may we add a test about this discussed case:
1) add a document `{"doc_id": 1, "content": "foo"}`
2) add a document `{"doc_id": "1", "content": "bar"}`
3) check if there is 2 different documents in DB
3) check if the content of the first document is "foo"

MarinPostma:
@ManyTheFish they should be replaced, like in my example :)

PR #380, opened 2021-09-30
Reserved keyword error message

irevoire:
And I missed _another_ reserved keyword error message in the filter :(

### DIFF
--- a/milli/src/asc_desc.rs
+++ b/milli/src/asc_desc.rs
@@ -20,12 +20,12 @@ impl fmt::Display for AscDescError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             Self::InvalidSyntax { name } => {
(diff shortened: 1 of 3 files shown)


irevoire:
I tested this in meilisearch and I get the expected result:
image

irevoire:
oops wait maybe we're going to change the message for `_geo`

irevoire:
Here are the new error messages:
image

And here is the spec patch:
https://github.com/meilisearch/specifications/pull/79/files

curquiza (dismissed):
I always feel a little bit lost in the middle of all these errors, but as far as I understand, it looks good to me 😇

irevoire:
I added `.` to the end of each error like in the spec but it's ready to be merged :+1:

Kerollmops (approved):
Thank you tamotamo!
bors merge

PR #383, opened 2021-10-07
Add check on latitude and longitude

irevoire:
Latitudes are not supposed to go beyond 90 degrees or below -90.
The same goes for longitudes with 180 or -180.

This was badly implemented in the filters, and was not implemented for the `AscDesc` rules.


irevoire:
bors merge

PR #409, opened 2021-11-03
remove update_id in UpdateBuilder

MarinPostma:
Removing the `update_id` from `UpdateBuidler`, since it serves no purpose. I had introduced it when working in HA some time ago, but I think there are better ways to do it now, so it can be removed an stop being in our way.

### DIFF
--- a/benchmarks/benches/indexing.rs
+++ b/benchmarks/benches/indexing.rs
@@ -39,7 +39,7 @@ fn indexing_songs_default(c: &mut Criterion) {
             move || {
                 let index = setup_index();
 
-                let update_builder = UpdateBuilder::new(0);
+                let update_builder = UpdateBuilder::new();
                 let mut wtxn = index.write_txn().unwrap();
                 let mut builder = update_builder.settings(&mut wtxn, &index);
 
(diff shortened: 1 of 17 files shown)


ManyTheFish:
Wait for [the PR on error message](https://github.com/meilisearch/milli/pull/412) to be merged, before merging this one.
Thanks! 🙏

milli/src/search/facet/filter_condition.rs
```diff
@@ -0,0 +1,929 @@
...
+        .unwrap();
+        let expected = FilterCondition::from_str(
+            &rtxn,
+            &index,
+            "channel = gotaga AND (timestamp = 44 OR channel != ponce)",
+        )
+        .unwrap();
+        assert_eq!(condition, expected);
+    }
+}
```
ManyTheFish, line 929: This file is not imported in milli, you may not want to add it to this PR.
Moreover, it seems unrelated.

ManyTheFish (changes requested):

milli/src/search/facet/filter_condition.rs
MarinPostma: uhm the rebase must have picked it up

ManyTheFish (approved):
bors merge

PR #416, opened 2021-11-18
Update tokenizer v0.2.6

ManyTheFish:
- Update tokenizer v0.2.6
- Export tokenizer from milli in order to use the same version between milli and Meilisearch


Kerollmops (approved):
I think that's valid PR!
bors merge

PR #418, opened 2021-11-22
change visibility of DocumentDeletionResult

MarinPostma:
Change the visibility of `DocumentDeletionResult`, so its fields can be accesses from outside milli.


Kerollmops (dismissed):
Ho yeah, strange that rustc didn't warn about exposing a private type from a public function.
bors merge

MarinPostma:
@Kerollmops I fixed fmt :)

Kerollmops (approved):
bors merge

PR #422, opened 2021-12-09
Prefer returning `None` instead of using an `FilterCondition::Empty` state

Kerollmops:
This PR is related to the issue comment https://github.com/meilisearch/MeiliSearch/issues/1338#issuecomment-989322889 which exhibits the fact that when a filter is known to be empty no results are returned which is wrong, the filter should not apply as no restriction is done on the documents set.

The filter system on the milli side has introduced an Empty state which was used in this kind of situation but I found out that it is not needed and that when we parse a filter and that it is empty we can simply return `None` as the `Filter::from_array` constructor does. So I removed it and added tests!

On the MeiliSearch side, we just need to match on a `None` and completely ignore the filter in such a case.

### DIFF
--- a/benchmarks/benches/utils.rs
+++ b/benchmarks/benches/utils.rs
@@ -117,7 +117,7 @@ pub fn run_benches(c: &mut criterion::Criterion, confs: &[Conf]) {
                     let mut search = index.search(&rtxn);
                     search.query(query).optional_words(conf.optional_words);
                     if let Some(filter) = conf.filter {
-                        let filter = Filter::from_str(filter).unwrap();
+                        let filter = Filter::from_str(filter).unwrap().unwrap();
                         search.filter(filter);
                     }
                     if let Some(sort) = &conf.sort {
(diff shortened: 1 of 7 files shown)


bidoubiwa:
I answered on the related issue with what I think raises my point. https://github.com/meilisearch/MeiliSearch/issues/1338#issuecomment-989322889
With this PR it fixes the problem I raised?

curquiza:
@meilisearch/product-team can you confirm this is the behavior you expect?

Kerollmops:
In fact, this is an issue that seems to have been raised when we switch from pest to nom, we used the `Filter::Empty` variant instead of returning an optional which breaks this behavior. It should never have behaved like this and don't restrict the results at all instead of returning nothing when the filter is empty.

irevoire (approved):

Kerollmops:
Thank you for the review!
bors merge

gmourier:
> @meilisearch/product-team can you confirm this is the behavior you expect?

This is consistent with the behavior of other fields like `attributesToRetrieve`which can be sent as `[]` for example. Since we have several possible filter syntaxes, it makes sense to be able to specify "" as `[]` for this field.

PR #424, opened 2021-12-14
Store the geopoint in three dimensions

irevoire:
Related to this issue: https://github.com/meilisearch/MeiliSearch/issues/1872

Fix the whole computation of distance for any “geo” operations (sort or filter). Now when you sort points they are returned to you in the right order.
And when you filter on a specific radius you only get points included in the radius.

This PR changes the way we store the geo points in the RTree.
Instead of considering the latitude and longitude as orthogonal coordinates, we convert them to real orthogonal coordinates projected on a sphere with a radius of 1.
This is the conversion formulae.
image
Which, in rust, translate to this function:
```rust
pub fn lat_lng_to_xyz(coord: &[f64; 2]) -> [f64; 3] {
    let [lat, lng] = coord.map(|f| f.to_radians());
    let x = lat.cos() * lng.cos();
    let y = lat.cos() * lng.sin();
    let z = lat.sin();

    [x, y, z]
}
```

Storing the points on a sphere is easier / faster to compute than storing the point on an approximation of the real earth shape.
But when we need to compute the distance between two points we still need to use the haversine distance which works with latitude and longitude.
So, to do the fewest search-time computation possible I'm now associating every point with its `DocId` and its lat/lng.

### DIFF
--- a/milli/src/lib.rs
+++ b/milli/src/lib.rs
@@ -54,7 +54,11 @@ pub type FieldId = u16;
 pub type Position = u32;
 pub type RelativePosition = u16;
 pub type FieldDistribution = BTreeMap<String, u64>;
-pub type GeoPoint = rstar::primitives::GeomWithData<[f64; 2], DocumentId>;
+
+/// A GeoPoint is a point in cartesian plan, called xyz_point in the code. Its metadata
+/// is a tuple composed of 1. the DocumentId of the associated document and 2. the original point
+/// expressed in term of latitude and longitude.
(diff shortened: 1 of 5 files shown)


curquiza:
⚠️ Should be merged once MeiliSearch v0.25.0 is released :) This fix is for v0.26.0

Kerollmops (approved):
Hope the RTree works better with 3D points, will merge this in v0.26.

Kerollmops (approved):
bors merge

PR #429, opened 2022-01-04
Benchmark CIs: not use a default label to call the GH runner

curquiza:
Since we now have multiple self-hosted github runners, we need to differentiate them calling them in the CI. The `self-hosted` label is the default one, so we need to use the unique and appropriate one for the benchmark machine

Capture d’écran 2022-01-04 à 15 42 18


irevoire (approved):
:+1:

irevoire:
bors merge

PR #430, opened 2022-01-11
Document batch support

MarinPostma:
This pr adds support for document batches in milli. It changes the API of the `IndexDocuments` builder by adding a `add_documents` method. The API of the updates is changed a little, with the `UpdateBuilder` being renamed to `IndexerConfig` and being passed to the update builders. This makes it easier to pass around structs that need to access the indexer config, rather that extracting the fields each time. This change impacts many function signatures and simplify them.

The change in not thorough, and may require another PR to propagate to the whole codebase. I restricted to the necessary for this PR.


milli/src/update/index_documents/mod.rs
```diff
@@ -55,120 +51,115 @@ pub enum IndexDocumentsMethod {
...
-    pub fn execute<R, F>(
-        self,
-        reader: DocumentBatchReader<R>,
-        progress_callback: F,
-    ) -> Result<DocumentAdditionResult>
+    /// Adds a batch of documents to the current builder.
+    ///
+    /// Since the documents are progressively added to the writer, a failure will cause a stale
+    /// builder, and the builder must be discarded.
+    #[must_use = "a builder must not be used after a add document operation reported an error."]
```
Kerollmops, line 121: It is not how `must_use` works, [here it is advertising the user that he is not using the u64 output by this method](https://rust-lang.github.io/rfcs/1940-must-use-functions.html).

Kerollmops: Can you add a point about what this return `u64` is, please?

milli/src/update/index_documents/transform.rs
```diff
@@ -84,56 +81,73 @@ fn find_primary_key(index: &DocumentsBatchIndex) -> Option<&str> {
         .map(String::as_str)
 }
 
-impl Transform<'_, '_> {
+impl<'i, 'a> Transform<'a, 'i> {
```
Kerollmops, line 84:
```suggestion
impl<'a, 'i> Transform<'a, 'i> {
```

milli/src/update/index_documents/transform.rs
```diff
@@ -230,38 +244,40 @@ impl Transform<'_, '_> {
             documents_seen: documents_count,
         });
 
+        self.index.put_fields_ids_map(wtxn, &fields_ids_map)?;
+        self.index.put_primary_key(wtxn, &primary_key_name)?;
```
Kerollmops, line 248: Ok, now the transform struct writes into the database, to keep a constant state? For example the new fields it found when iterating through documents?

milli/src/update/settings.rs
```diff
@@ -118,14 +110,10 @@ impl<'a, 't, 'u, 'i> Settings<'a, 't, 'u, 'i> {
...
         }
     }
 
-    pub fn log_every_n(&mut self, n: usize) {
-        self.log_every_n = Some(n);
-    }
-
     pub fn reset_searchable_fields(&mut self) {
         self.searchable_fields = Setting::Reset;
     }
```
Kerollmops, line 119: Why don't you remove those functions too?

milli/src/update/update_builder.rs
```diff
@@ -1,130 +0,0 @@
-use grenad::CompressionType;
```
Kerollmops, line 1: Have you deleted this file?

Kerollmops (changes requested):

milli/src/update/update_builder.rs
MarinPostma: it was renamed to `indexer_config.rs`, but git hasn't picked it up

milli/src/update/settings.rs
MarinPostma: they are not part of the `IndexerConfig`, we could create another struct to hold them, and have it passed to the `Settings` constructor, like we do for the `IndexDocumentsConfig`, but that seemed out of scope for this PR.

milli/src/update/index_documents/transform.rs
MarinPostma: that's right

milli/src/update/settings.rs
Kerollmops: Good point!

Kerollmops (changes requested):
Looks good to me, you should do a rustfmt pass :)

Kerollmops (dismissed):

MarinPostma:
bors try

Kerollmops:
Ok, do we want to merge that now? I would prefer that we wait for the fix of @irevoire about https://github.com/meilisearch/MeiliSearch/issues/2082 first.

MarinPostma:
@Kerollmops, I think we can merge that now

Kerollmops:
Yup, indeed!
bors merge

MarinPostma:
bors try

Kerollmops:
bors merge

Kerollmops (approved):
My bad 🤦‍♂️
bors merge

PR #431, opened 2022-01-12
Fix and improve word prefix pair proximity

Kerollmops:
This PR first fixes the algorithm we used to select and compute the word prefix pair proximity database. The previous version was skipping nearly all of the prefixes. The issue is that this fix made this method to take more time and we were trying to reduce the time spent in it.

With @ManyTheFish we found out that we could skip some of the work we were doing by:
 - discarding the prefixes that were shorter than a specific threshold (default: 2).
 - discarding the word prefix pairs with proximity bigger than a specific threshold (default: 4).
 - remove the unused threshold that was specifying a minimum amount of word docids to merge.

We will take more time to do some more optimization, like stop clearing and recomputing from scratch the database, we will compute the subsets of keys to create, keep and merge. This change is a little bit more complex than what this PR does.

I keep this PR as a draft as I want to further test the real gain if it is enough or not if it is valid or not. I advise reviewers to review commit by commit to see the changes bit by bit, reviewing the whole PR can be hard.

### DIFF
--- a/milli/src/update/word_prefix_pair_proximity_docids.rs
+++ b/milli/src/update/word_prefix_pair_proximity_docids.rs
@@ -19,6 +19,7 @@ pub struct WordPrefixPairProximityDocids<'t, 'u, 'i> {
     pub(crate) max_nb_chunks: Option<usize>,
     pub(crate) max_memory: Option<usize>,
     max_proximity: u8,
+    max_prefix_length: usize,
 }
 
 impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
@@ -34,6 +35,7 @@ impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
             max_nb_chunks: None,
             max_memory: None,
             max_proximity: 4,
+            max_prefix_length: 2,
         }
     }
 
@@ -45,6 +47,17 @@ impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
     /// to this bound otherwise.
     pub fn max_proximity(&mut self, value: u8) -> &mut Self {
         self.max_proximity = value.max(7);
+        self
+    }
+
+    /// Set the maximum length the prefix of a word pair is allowed to have be part of the words
+    /// prefixes database. If two words are two far from the threshold the associated documents
+    /// will not be part of the prefix database.
+    ///
+    /// Default value is 4. This value must be lower or equal than 4 and will be clamped
+    /// to this bound otherwise.
+    pub fn max_prefix_length(&mut self, value: usize) -> &mut Self {
+        self.max_prefix_length = value;
         self
     }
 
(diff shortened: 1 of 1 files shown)


milli/src/update/word_prefix_pair_proximity_docids.rs
```diff
@@ -33,18 +34,30 @@ impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
...
+        self.max_proximity = value.max(7);
+        self
+    }
+
+    /// Set the maximum length the prefix of a word pair is allowed to have be part of the words
+    /// prefixes database. If two words are two far from the threshold the associated documents
+    /// will not be part of the prefix database.
+    ///
+    /// Default value is 4. This value must be lower or equal than 4 and will be clamped
+    /// to this bound otherwise.
```
ManyTheFish, line 58:
```suggestion
    /// Set the maximum length the prefix of a word pair is allowed to have to be part of the words
    /// prefixes database. If the prefix length is higher than the threshold, the associated documents
    /// will not be part of the prefix database.
    ///
    /// Default value is 2.
```

milli/src/update/word_prefix_pair_proximity_docids.rs
```diff
@@ -33,18 +34,30 @@ impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
...
+            max_prefix_length: 2,
         }
     }
 
-    /// Set the number of words required to make a prefix be part of the words prefixes
-    /// database. If a word prefix is supposed to match more than this number of words in the
-    /// dictionnary, therefore this prefix is added to the words prefixes datastructures.
+    /// Set the maximum proximity required to make a prefix be part of the words prefixes
+    /// database. If two words are two far from the threshold the associated documents will
+    /// not be part of the prefix database.
```
ManyTheFish, line 44:
```suggestion
    /// Set the maximum proximity required to make a prefix be part of the words prefixes
    /// database. If two words are too far from the threshold the associated documents will
    /// not be part of the prefix database.
```

milli/src/update/word_prefix_pair_proximity_docids.rs
```diff
@@ -33,18 +34,30 @@ impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
...
     ///
-    /// Default value is 100. This value must be higher than 50 and will be clamped
-    /// to these bound otherwise.
-    pub fn threshold(&mut self, value: u32) -> &mut Self {
-        self.threshold = value.max(50);
+    /// Default value is 4. This value must be lower or equal than 4 and will be clamped
+    /// to this bound otherwise.
```
ManyTheFish, line 47:
```suggestion
    /// Default value is 4. This value must be lower or equal than 7 and will be clamped
    /// to this bound otherwise.
```

milli/src/update/word_prefix_pair_proximity_docids.rs
```diff
@@ -64,44 +77,47 @@ impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
...
-                    buffer.truncate(w1.len() + 1);
-                    buffer.extend_from_slice(prefix);
-                    buffer.push(prox);
-
-                    match prefixes_cache.get_mut(&buffer) {
-                        Some(value) => value.push(data),
-                        None => {
-                            prefixes_cache.insert(buffer.clone(), vec![data]);
+                for prefix in prefixes.iter() {
+                    if prefix.len() <= self.max_prefix_length && w2.starts_with(prefix) {
```
ManyTheFish, line 111: the condition `prefix.len() <= self.max_prefix_length` has probably no cost compared to the rest, but, shouldn't we prefilter prefixes coming from the fst before beginning the computation?

ManyTheFish (changes requested):

milli/src/update/word_prefix_pair_proximity_docids.rs
Kerollmops: Do you mean that it would be more logical to change the order of both conditions in this if?

ManyTheFish: no, this condition could be removed from this line being done before the double loop line 88:
```rust
// filter here on prefix length.
while let Some(((w1, w2, prox), data)) = db.next().transpose()? {
                // [...]
                for prefix in prefixes.iter() {
                    // instead of here.
                    if prefix.len() <= self.max_prefix_length && w2.starts_with(prefix) {
                        // [...]
                    }
               }
}
```

ManyTheFish: I don't know if it is possible, but, can you filter directly in the FST the prefixes that have a length under `self.max_prefix_length`

Kerollmops: It is possible to filter out the prefixes by their lengths when we store them in the `prefix_fst_keys` vec but it would imply that we create a sub-vec with all of the filtered values. Now we just have a simple slice that we receive from `slice_group_by`. I prefer to keep it like that for now.

ManyTheFish (approved):
Bors merge

PR #432, opened 2022-01-13
Fuzzer

irevoire:
Provide a first way of fuzzing the indexing part of milli.
It depends on [cargo-fuzz](https://rust-fuzz.github.io/book/cargo-fuzz.html)

### DIFF
--- a/milli/fuzz/fuzz_targets/indexing.rs
+++ b/milli/fuzz/fuzz_targets/indexing.rs
@@ -1,0 +1,76 @@
+#![no_main]
+
+use std::io::{BufWriter, Cursor, Read, Seek, Write};
+
+use anyhow::{bail, Result};
+use arbitrary_json::ArbitraryValue;
+use heed::EnvOpenOptions;
+use libfuzzer_sys::fuzz_target;
+use milli::documents::{DocumentBatchBuilder, DocumentBatchReader};
+use milli::update::UpdateBuilder;
+use milli::Index;
+use serde_json::Value;
+
+#[cfg(target_os = "linux")]
+#[global_allocator]
+static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
+
+/// reads json from input and write an obkv batch to writer.
+pub fn read_json(input: impl Read, writer: impl Write + Seek) -> Result<usize> {
+    let writer = BufWriter::new(writer);
+    let mut builder = DocumentBatchBuilder::new(writer)?;
+    builder.extend_from_json(input)?;
+
+    if builder.len() == 0 {
+        bail!("Empty payload");
+    }
+
+    let count = builder.finish()?;
+
+    Ok(count)
+}
+
+fn index_documents(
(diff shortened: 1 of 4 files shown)


milli/fuzz/fuzz_targets/indexing.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    let faceted_fields: HashSet<String> =
+        ["released-timestamp", "duration-float", "genre", "country", "artist"]
+            .iter()
+            .map(|s| s.to_string())
+            .collect();
+    builder.set_filterable_fields(faceted_fields.clone());
+    builder.set_sortable_fields(faceted_fields);
+
+    builder.set_distinct_field("same".to_string());
```
Kerollmops, line 76: What's "same" here? this doesn't exist, right?

milli/fuzz/fuzz_targets/indexing.rs
```diff
@@ -0,0 +1,106 @@
...
+
+            let mut documents = Cursor::new(Vec::new());
+
+            // We ignore all badly generated documents
+            if let Ok(_count) = read_json(json.as_bytes(), &mut documents) {
+                documents.rewind().unwrap();
+                let documents = DocumentBatchReader::from_reader(documents).unwrap();
+                match index_documents(&mut index, documents) {
+                    _ => (),
+                }
```
Kerollmops, line 100: Why do you match on nothing here?

milli/fuzz/fuzz_targets/indexing.rs
```diff
@@ -0,0 +1,106 @@
...
+        for batch in batches {
+            let documents: Vec<Value> =
+                batch.into_iter().map(|value| serde_json::Value::from(value)).collect();
+            let json = Value::Array(documents);
+            let json = serde_json::to_string(&json).unwrap();
+
+            let mut documents = Cursor::new(Vec::new());
+
+            // We ignore all badly generated documents
+            if let Ok(_count) = read_json(json.as_bytes(), &mut documents) {
```
Kerollmops, line 95: Don't you want to catch possible issues, here?

Kerollmops (changes requested):

milli/fuzz/fuzz_targets/indexing.rs
irevoire: Yes, we don't know what will exist since it's the fuzzer who's going to generate the dataset.

milli/fuzz/fuzz_targets/indexing.rs
irevoire: Because at first I wanted to catch some specific error on milli but finally I didn't know what we could catch and just removed everything, thus I'm only looking for unexpected panic!
I'm updating the code thanks :+1:

Kerollmops (approved):
We can merge that, thank you. Note that this fuzzer must be run by hand.
bors merge

PR #433, opened 2022-01-17
fix(filter): Fix two bugs.

irevoire:
- Stop lowercasing the field when looking in the field id map
- When a field id does not exist it means there is currently zero
  documents containing this field thus we return an empty RoaringBitmap
  instead of throwing an internal error

Will fix https://github.com/meilisearch/MeiliSearch/issues/2082 once meilisearch is released

### DIFF
--- a/milli/src/search/facet/filter.rs
+++ b/milli/src/search/facet/filter.rs
@@ -33,7 +33,6 @@ enum FilterError<'a> {
     BadGeoLng(f64),
     Reserved(&'a str),
     TooDeep,
-    InternalError,
 }
 impl<'a> std::error::Error for FilterError<'a> {}
 
@@ -58,7 +57,6 @@ impl<'a> Display for FilterError<'a> {
             Self::BadGeo(keyword) => write!(f, "`{}` is a reserved keyword and thus can't be used as a filter expression. Use the _geoRadius(latitude, longitude, distance) built-in rule to filter on _geo field coordinates.", keyword),
             Self::BadGeoLat(lat) => write!(f, "Bad latitude `{}`. Latitude must be contained between -90 and 90 degrees. ", lat),
             Self::BadGeoLng(lng) => write!(f, "Bad longitude `{}`. Longitude must be contained between -180 and 180 degrees. ", lng),
-            Self::InternalError => write!(f, "Internal error while executing this filter."),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


milli/src/search/facet/filter.rs
```diff
@@ -342,12 +340,12 @@ impl<'a> Filter<'a> {
         match &self.condition {
             FilterCondition::Condition { fid, op } => {
                 let filterable_fields = index.filterable_fields(rtxn)?;
-                if filterable_fields.contains(&fid.to_lowercase()) {
+                if filterable_fields.contains(fid.value()) {
```
Kerollmops, line 343: Don't you implement `Deref` on that type? Do you need to call `value()` here?

milli/src/search/facet/filter.rs
```diff
@@ -58,7 +57,6 @@ impl<'a> Display for FilterError<'a> {
             Self::BadGeo(keyword) => write!(f, "`{}` is a reserved keyword and thus can't be used as a filter expression. Use the _geoRadius(latitude, longitude, distance) built-in rule to filter on _geo field coordinates.", keyword),
             Self::BadGeoLat(lat) => write!(f, "Bad latitude `{}`. Latitude must be contained between -90 and 90 degrees. ", lat),
             Self::BadGeoLng(lng) => write!(f, "Bad longitude `{}`. Longitude must be contained between -180 and 180 degrees. ", lng),
-            Self::InternalError => write!(f, "Internal error while executing this filter."),
```
Kerollmops, line 61: Ho! I'm glad we remove this variant! 😃

milli/src/search/facet/filter.rs
```diff
@@ -342,12 +340,12 @@ impl<'a> Filter<'a> {
...
                 let filterable_fields = index.filterable_fields(rtxn)?;
-                if filterable_fields.contains(&fid.to_lowercase()) {
+                if filterable_fields.contains(fid.value()) {
                     let field_ids_map = index.fields_ids_map(rtxn)?;
-                    if let Some(fid) = field_ids_map.id(&fid) {
+                    if let Some(fid) = field_ids_map.id(fid.value()) {
                         Self::evaluate_operator(rtxn, index, numbers_db, strings_db, fid, &op)
                     } else {
-                        return Err(fid.as_external_error(FilterError::InternalError))?;
+                        return Ok(RoaringBitmap::new());
```
Kerollmops, line 348: Do you think this is absolutely valid, though? That when a field can't be found in the `FieldsIdsMap` but has effectively been declared in the `filterableAttributes` it means that no document has ever been declared with this field and therefore we can return an empty set?

milli/src/search/facet/filter.rs
irevoire: Yes but `Deref` returns the « raw » value and I think here it's better to get the real value in it (with the unescaped quote if there was any)

milli/src/search/facet/filter.rs
irevoire: Same :joy:

milli/src/search/facet/filter.rs
Kerollmops: Ho, yeah. Not sure if Deref should return the raw or escaped value here :) But that'll part of another PR 😄

milli/src/search/facet/filter.rs
irevoire: I think it's valid yes

Kerollmops (approved):
Thank you for the fix, it looks valid!
bors merge

PR #435, opened 2022-01-18
Ensure we get no documents and no error when filtering on an empty db

irevoire:


Kerollmops (approved):
Nice, I approve that!
bors merge

PR #436, opened 2022-01-18
Speed up the word prefix databases computation time

Kerollmops:
This PR depends on the fixes done in #431 and must be merged after it.

In this PR we will bring the `WordPrefixPairProximityDocids`, `WordPrefixDocids` and, `WordPrefixPositionDocids` update structures to a new era, a better era, where computing the word prefix pair proximities costs much fewer CPU cycles, an era where this update structure can use the, previously computed, set of new word docids from the newly indexed batch of documents.

---

The `WordPrefixPairProximityDocids` is an update structure, which means that it is an object that we feed with some parameters and which modifies the LMDB database of an index when asked for. This structure specifically computes the list of word prefix pair proximities, which correspond to a list of pairs of words associated with a proximity (the distance between both words) where the second word is not a word but a prefix e.g. `s`, `se`, `a`. This word prefix pair proximity is associated with the list of documents ids which contains the pair of words and prefix at the given proximity.

The origin of the performances issue that this struct brings is related to the fact that it starts its job from the beginning, it clears the LMDB database before rewriting everything from scratch, using the other LMDB databases to achieve that. I hope you understand that this is absolutely not an optimized way of doing things.

### DIFF
--- a/milli/src/update/index_documents/mod.rs
+++ b/milli/src/update/index_documents/mod.rs
@@ -285,6 +285,7 @@ where
         let index_is_empty = index_documents_ids.len() == 0;
         let mut final_documents_ids = RoaringBitmap::new();
         let mut word_pair_proximity_docids = Vec::new();
+        let mut word_docids = Vec::new();
 
         let mut databases_seen = 0;
         (self.progress)(UpdateIndexingStep::MergeDataIntoFinalDatabase {
@@ -294,6 +295,19 @@ where
 
         for result in lmdb_writer_rx {
             let typed_chunk = match result? {
+                TypedChunk::WordDocids(chunk) => {
+                    // We extract and mmap our chunk file to be able to get it for next processes.
+                    let mut file = chunk.into_inner();
+                    let mmap = unsafe { memmap2::Mmap::map(&file)? };
+                    let cursor_mmap = CursorClonableMmap::new(ClonableMmap::from(mmap));
+                    let chunk = grenad::Reader::new(cursor_mmap)?;
+                    word_docids.push(chunk);
+
+                    // We reconstruct our typed-chunk back.
+                    file.rewind()?;
+                    let chunk = grenad::Reader::new(file)?;
+                    TypedChunk::WordDocids(chunk)
+                }
                 TypedChunk::WordPairProximityDocids(chunk) => {
                     // We extract and mmap our chunk file to be able to get it for next processes.
                     let mut file = chunk.into_inner();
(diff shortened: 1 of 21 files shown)


milli/src/update/index_documents/mod.rs
```diff
@@ -282,16 +279,62 @@ where
...
+                    // We extract and mmap our chunk file to be able to get it for next processes.
+                    let mut file = chunk.into_inner();
+                    let mmap = unsafe { memmap2::Mmap::map(&file)? };
+                    let cursor_mmap = CursorClonableMmap::new(ClonableMmap::from(mmap));
+                    let chunk = grenad::Reader::new(cursor_mmap)?;
+                    word_docids.push(chunk);
+
+                    // We reconstruct our typed-chunk back.
+                    file.rewind()?;
+                    let chunk = grenad::Reader::new(file)?;
```
ManyTheFish, line 304: could you factorize this in a function?
Is [into_clonable_grenad](https://github.com/meilisearch/milli/blob/main/milli/src/update/index_documents/helpers/grenad_helpers.rs#L69-L77) could suit this case?

ManyTheFish (changes requested):

curquiza:
Is it breaking? (API break or DB break)

milli/src/update/index_documents/mod.rs
Kerollmops: Done by bumping the grenad version to v0.4.1 and then changing the behavior of the `into_cloneable_grenad` (renamed `as_cloneable_grenad`) by naming it take a reference of a `grenad::Reader`. Doing so reduced the complexity of this quoted code part.

Kerollmops:
> Is this PR breaking? (API break or DB break)

There shouldn't be any breaking changes either on the Rust API or the written formats here. If there are any it's a bug.

Kerollmops:
I'll merge then, thank you for your review!
bors merge

milli/src/update/index_documents/mod.rs
```diff
@@ -363,6 +397,27 @@ where
...
+        // We retrieve the newly added words between the previous and new prefix word fst.
+        let new_prefix_fst_words = fst_stream_into_vec(
+            current_prefix_fst.op().add(&previous_words_prefixes_fst).difference(),
+        );
+
+        // We compute the set of prefixes that are no more part of the prefix fst.
+        let del_prefix_fst_words = fst_stream_into_hashset(
+            previous_words_prefixes_fst.op().add(&current_prefix_fst).difference(),
+        );
+
```
Kerollmops, line 420: We must stop creating enormous `Vec`s when we process the first update. In this code snippet, on an initial update, we compute the difference between the previous prefix fst (empty) and the current prefix fst (big) and store the result in a `Vec` and an `HashSet` (takes time and space).

PR #438, opened 2022-01-21
CLI improvements

MarinPostma:
I've made the following changes to the cli:
- `settings-update` become `settings`, with two subcommands: `update` and `show`.
- `document-addition` becomes `documents` with a subcommands: `add` (I'll add a feature to list documents later)
- `search` now has an interactive mode `-i`
- search return the number of documents and the time it took to perform the search.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1,7 +1,9 @@
+use std::collections::BTreeMap;
 use std::fs::File;
-use std::io::{stdin, BufRead, BufReader, Cursor, Read};
+use std::io::{stdin, BufRead, BufReader, Cursor, Read, Write};
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::time::Instant;
 
 use byte_unit::Byte;
 use eyre::Result;
@@ -10,6 +12,8 @@ use milli::update::UpdateIndexingStep::{
     ComputeIdsAndMergeDocuments, IndexDocuments, MergeDataIntoFinalDatabase, RemapDocumentAddition,
 };
 use milli::update::{IndexDocumentsConfig, IndexDocumentsMethod, IndexerConfig};
+use milli::Index;
+use serde_json::{Map, Value};
 use structopt::StructOpt;
 
 #[cfg(target_os = "linux")]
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -32,9 +36,108 @@ struct Cli {
 
 #[derive(Debug, StructOpt)]
 enum Command {
-    DocumentAddition(DocumentAddition),
+    Documents {
+        #[structopt(subcommand)]
+        cmd: Documents,
+    },
```
Kerollmops, line 42:
```suggestion
    #[structopt(subcommand)]
    Documents(Documents),
```

cli/src/main.rs
```diff
@@ -32,9 +36,108 @@ struct Cli {
...
     Search(Search),
-    SettingsUpdate(SettingsUpdate),
+    Settings {
+        #[structopt(subcommand)]
+        cmd: Settings,
+    },
```
Kerollmops, line 47:
```suggestion
    #[structopt(subcommand)]
    Settings(Settings),
```

cli/src/main.rs
```diff
@@ -32,9 +36,108 @@ struct Cli {
...
+}
+
+impl Performer for Command {
+    fn perform(self, index: Index) -> Result<()> {
+        match self {
+            Command::Documents { cmd } => cmd.perform(index),
+            Command::Search(cmd) => cmd.perform(index),
+            Command::Settings { cmd } => cmd.perform(index),
```
Kerollmops, line 55:
```suggestion
            Command::Documents(cmd) => cmd.perform(index),
            Command::Search(cmd) => cmd.perform(index),
            Command::Settings(cmd) => cmd.perform(index),
```

cli/src/main.rs
```diff
@@ -247,29 +346,83 @@ struct Search {
     offset: Option<usize>,
     #[structopt(short, long)]
     limit: Option<usize>,
+    #[structopt(short, long, conflicts_with_all=&["query", "filter", "offset", "limit"])]
```
Kerollmops, line 349:
```suggestion
    #[structopt(short, long, conflicts_with = "query")]
```

Why would it be invalid to specify filters, offsets, or limits?

cli/src/main.rs
```diff
@@ -247,29 +346,83 @@ struct Search {
...
+            let stdin = std::io::stdin();
+            let mut lines = stdin.lock().lines();
+            loop {
+                eprint!("> ");
+                std::io::stdout().flush()?;
+                match lines.next() {
+                    Some(Ok(line)) => {
+                        let now = Instant::now();
+                        let jsons =
+                            Self::perform_single_search(&index, &Some(line), &None, &None, &None)?;
```
Kerollmops, line 365:
```suggestion
                            Self::perform_single_search(&index, &Some(line), &self.filter, &self.offset, &self.limit)?;
```

Kerollmops (changes requested):

cli/src/main.rs
MarinPostma: Because it is not yet implemented

MarinPostma: My goal was to enable it from the interactive prompt instead, so those will always be ignored

cli/src/main.rs
MarinPostma: this is not permitted by structopt

Kerollmops (approved):
Looks good to me!
bors merge

PR #440, opened 2022-01-25
fix(fuzzer): fix the fuzzer after #430

irevoire:


Kerollmops (approved):
Looks good to me!
bors merge

PR #442, opened 2022-02-01
fix phrase search

MarinPostma:
This should fix phrase search false positive. Instead of looking at words in a phrase search pairwise, we take the first word as reference, and we check earch word to see if they are at the right distance to the first word.

For example, with `known to your people`:
- known is the baseline
- search `known` and `to` with proximity `1`
- search `known` and `your` with proximity `2`
- search `known` and `people` with proximity `3`

intersect the documents that satisfy those conditions.


milli/src/search/criteria/mod.rs
```diff
@@ -318,21 +319,41 @@ pub fn resolve_query_tree<'t>(
...
+                    for bitmap in bitmaps {
+                        if first_iter {
+                            candidates = bitmap;
+                            first_iter = false;
+                        } else {
+                            candidates &= bitmap;
+                        }
+                        // There will be no match, return early
+                        if candidates.is_empty() {
+                            break;
```
ManyTheFish, line 355: why don't you move this out of the `for` loop Line 325? Making all the intersections at the same time.

milli/src/search/criteria/mod.rs
```diff
@@ -318,21 +319,41 @@ pub fn resolve_query_tree<'t>(
...
+                        .map(|((pos1, s1), (pos2, s2))| (s1, s2, pos2 - pos1));
+
+                    let mut bitmaps = Vec::with_capacity(winsize.pow(2));
+
+                    for (s1, s2, d) in dists {
+                        match ctx.word_pair_proximity_docids(s1, s2, d as u8)? {
+                            Some(m) => bitmaps.push(m),
+                            None => return Ok(RoaringBitmap::new()),
+                        }
+                    }
```
ManyTheFish, line 341: What do you think about something like that ? 🤔 
```suggestion
                for mut win in words.windows(winsize) {
                    let mut bitmaps = Vec::with_capacity(winsize.pow(2));

                    while let Some((head, tail)) = win.split_first() {
                        for (dist, w) in tail.iter().enumerate() {
                            match ctx.word_pair_proximity_docids(s1, s2, dist as u8 + 1)? {
                                Some(m) => bitmaps.push(m),
                                None => return Ok(RoaringBitmap::new()),
                            }
                        }

                        win = tail;
                    }
```

ManyTheFish (changes requested):

milli/src/search/criteria/mod.rs
MarinPostma: unless you have strong reasons, I would favour the declarative approach

milli/src/search/criteria/mod.rs
MarinPostma: This would only be useful for phrase with more than 8 words, but the second window iteration will contains mostly bitmap fetch that we have already seen. This is very wasteful in case there are no matches, so we first try to find matches. Also kero didn't want to put cache here for now.

milli/src/search/criteria/mod.rs
ManyTheFish: I find it shorter, this avoids doing 3 loops and a cartesian product with a filter is not so clear.
But I understand that the first cartesian product has "no cost" and the comment explains its purpose.

Kerollmops: Indeed, I find the Many's solution easier to understand and expressing the behavior better than having to understand that the Cartesian product is not used as is and values are filtered out afterwards.

MarinPostma: wdyt about that: 
```rust
                    let mut bitmaps = Vec::with_capacity(winsize.pow(2));
                    for (offset, s1) in win.iter().enumerate() {
                        for (dist, s2) in win.iter().skip(offset + 1).enumerate() {
                            match ctx.word_pair_proximity_docids(s1, s2, dist as u8 + 1)? {
                                Some(m) => bitmaps.push(m),
                                None => return Ok(RoaringBitmap::new()),
                            }
                        }
                    }
```

we avoid the mutable window this way, which is a bit awkward.

ManyTheFish: @MarinPostma, yes, this version is the clearest.
Meanwhile, I have a rust question, does the implementation `win.iter().skip(offset + 1).enumerate()` is the same of `win[offset + 1..].into_iter().enumerate()` in term of time?

MarinPostma: yes, but the former can't panic

milli/src/search/criteria/mod.rs
```diff
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
 use std::collections::HashMap;
 
+use itertools::Itertools;
```
Kerollmops, line 4:
```suggestion
```

Kerollmops (changes requested):

Kerollmops (approved):
Looks good to me, waiting for the @ManyTheFish review before merging!

ManyTheFish (approved):

PR #444, opened 2022-02-02
Fix the parsing of ndjson requests to index more than the first line

Kerollmops:
This PR correctly uses the `BufRead` trait to read every line of the content instead of just the first one. This bug was only affecting the http-ui test crate.


MarinPostma (approved):

Kerollmops:
bors merge

PR #445, opened 2022-02-03
allow null values in csv

MarinPostma:
This pr allows null values in csv:
- if the field is of type string, then an empty field is considered null (`,,`), anything other is turned into a string (i.e `, ,` is a single whitespace string)
- if the field is of type number, when the trimmed field is empty, we consider the value null (i.e `,,`, `, ,` are both null numbers) otherwise we try to parse the number.

### DIFF
--- a/milli/src/documents/builder.rs
+++ b/milli/src/documents/builder.rs
@@ -118,16 +118,26 @@ impl<W: io::Write + io::Seek> DocumentBatchBuilder<W> {
             for (value, (fid, ty)) in record.into_iter().zip(headers.iter()) {
                 let value = match ty {
                     AllowedType::Number => {
-                        value.parse::<f64>().map(Value::from).map_err(|error| {
-                            Error::ParseFloat {
-                                error,
-                                // +1 for the header offset.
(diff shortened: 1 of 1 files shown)


milli/src/documents/builder.rs
```diff
@@ -118,16 +118,26 @@ impl<W: io::Write + io::Seek> DocumentBatchBuilder<W> {
...
-                                error,
-                                // +1 for the header offset.
-                                line: i + 1,
-                                value: value.to_string(),
-                            }
-                        })?
+                        if value.trim().is_empty() {
+                            Value::Null
+                        } else {
+                            value.parse::<f64>().map(Value::from).map_err(|error| {
```
Kerollmops, line 124:
```suggestion
                            value.trim().parse::<f64>().map(Value::from).map_err(|error| {
```

According to our discution with @gmourier, we must trim before parsing the number.

Kerollmops (changes requested):

gmourier:
Thanks, @MarinPostma!

PR #456, opened 2022-02-28
Remove useless grenad merging

Kerollmops:


milli/src/update/index_documents/mod.rs
```diff
@@ -345,9 +345,9 @@ where
         self.index.put_documents_ids(self.wtxn, &all_documents_ids)?;
 
         self.execute_prefix_databases(
-            word_docids,
-            word_pair_proximity_docids,
-            word_position_docids,
+            word_docids.unwrap(),
+            word_pair_proximity_docids.unwrap(),
+            word_position_docids.unwrap(),
```
ManyTheFish, line 350: We may instead match this to compute or not each concerned prefix database, no?
something like:
```rust
if let Some(word_pair_proximity_docids) = word_pair_proximity_docids {
   // compute word prefix pair proximity database
}

Kerollmops: I prefer your solution indeed but I think we always generate the grenad files, they are just empty, am I right?

ManyTheFish:
@Kerollmops 
> @ManyTheFish are you sure I can skip merging the word_docids database?
yes, you can below the code that merges this database:
https://github.com/meilisearch/milli/blob/main/milli/src/update/index_documents/extract/mod.rs#L89-L97

ManyTheFish (dismissed):

ManyTheFish (approved):

Kerollmops:
bors merge

PR #458, opened 2022-03-01
Nested fields

irevoire:
For the following document:
```json
{
  "id": 1,
  "person": {
    "name": "tamo",
    "age": 25,
  }
}
```
Suppose the user sets `person` as a filterable attribute. We need to store `person` in the filterable _obviously_. But we also need to keep track of `person.name` and `person.age` somewhere.
That’s where I changed a little bit the logic of the engine.

Currently, we have a function called `faceted_field` that returns the union of the filterable and sortable.
I renamed this function in `user_defined_faceted_field`. And now, when we finish indexing documents, we look at all the fields and see if they « match » a `user_defined_faceted_field`.
So in our case:
- does `id` match `person`: 🔴 
- does `person.name` match `person`: 🟢 
- does `person.age` match `person`: 🟢 

And thus, we insert in the database the following faceted fields: `person, person.name, person.age`.

The good thing about that solution is that we generate everything during the indexing phase, and then during the search, we can access our field without recomputing too much globbing.

-----

Now the bad thing is that I had to create a new db.

And if that was only one db, that would be ok, but actually, I need to do the same for the:
- Displayed attributes
- Attributes to retrieve
- Attributes to highlight
- Attribute to crop

@Kerollmops 
Do you think there is a better way to do it?
Apart from all the code, can we have a problem because we have too many dbs?

### DIFF
--- a/milli/Cargo.toml
+++ b/milli/Cargo.toml
@@ -14,7 +14,7 @@ crossbeam-channel = "0.5.2"
 either = "1.6.1"
 fst = "0.4.7"
 fxhash = "0.2.1"
-flatten-serde-json = "0.1.0"
+flatten-serde-json = { path = "../flatten-serde-json" }
 grenad = { version = "0.4.1", default-features = false, features = ["tempfile"] }
 geoutils = "0.4.1"
 heed = { git = "https://github.com/meilisearch/heed", tag = "v0.12.1", default-features = false, features = ["lmdb", "sync-read-txn"] }
--- a/milli/Cargo.toml
+++ b/milli/Cargo.toml
@@ -14,6 +14,7 @@ crossbeam-channel = "0.5.2"
 either = "1.6.1"
 fst = "0.4.7"
 fxhash = "0.2.1"
+flatten-serde-json = "0.1.0"
 grenad = { version = "0.4.1", default-features = false, features = ["tempfile"] }
 geoutils = "0.4.1"
 heed = { git = "https://github.com/meilisearch/heed", tag = "v0.12.1", default-features = false, features = ["lmdb", "sync-read-txn"] }
--- a/milli/src/lib.rs
+++ b/milli/src/lib.rs
@@ -183,6 +183,43 @@ pub fn lat_lng_to_xyz(coord: &[f64; 2]) -> [f64; 3] {
     [x, y, z]
 }
 
+/// Returns `true` if the field match one of the faceted fields.
+/// See the function [`is_faceted_by`] below to see what “matching” means.
+pub fn is_faceted(field: &str, faceted_fields: impl IntoIterator<Item = impl AsRef<str>>) -> bool {
+    faceted_fields.into_iter().find(|facet| is_faceted_by(field, facet.as_ref())).is_some()
+}
+
+/// Returns `true` if the field match the facet.
+/// ```
+/// use milli::is_faceted_by;
+/// // -- the valid basics
+/// assert!(is_faceted_by("animaux", "animaux"));
+/// assert!(is_faceted_by("animaux.chien", "animaux"));
+/// assert!(is_faceted_by("animaux.chien.race.bouvier bernois.fourrure.couleur", "animaux"));
+/// assert!(is_faceted_by("animaux.chien.race.bouvier bernois.fourrure.couleur", "animaux.chien"));
+/// assert!(is_faceted_by("animaux.chien.race.bouvier bernois.fourrure.couleur", "animaux.chien.race.bouvier bernois"));
+/// assert!(is_faceted_by("animaux.chien.race.bouvier bernois.fourrure.couleur", "animaux.chien.race.bouvier bernois.fourrure"));
+/// assert!(is_faceted_by("animaux.chien.race.bouvier bernois.fourrure.couleur", "animaux.chien.race.bouvier bernois.fourrure.couleur"));
+///
+/// // -- the wrongs
+/// assert!(!is_faceted_by("chien", "chat"));
+/// assert!(!is_faceted_by("animaux", "animaux.chien"));
+/// assert!(!is_faceted_by("animaux.chien", "animaux.chat"));
+///
+/// // -- the strange edge cases
+/// assert!(!is_faceted_by("animaux.chien", "anima"));
+/// assert!(!is_faceted_by("animaux.chien", "animau"));
+/// assert!(!is_faceted_by("animaux.chien", "animaux."));
+/// assert!(!is_faceted_by("animaux.chien", "animaux.c"));
+/// assert!(!is_faceted_by("animaux.chien", "animaux.ch"));
+/// assert!(!is_faceted_by("animaux.chien", "animaux.chi"));
+/// assert!(!is_faceted_by("animaux.chien", "animaux.chie"));
+/// ```
+pub fn is_faceted_by(field: &str, facet: &str) -> bool {
+    field.starts_with(facet)
+        && field[facet.len()..].chars().next().map(|c| c == '.').unwrap_or(true)
+}
+
 #[cfg(test)]
 mod tests {
     use serde_json::json;
--- a/milli/src/update/index_documents/mod.rs
+++ b/milli/src/update/index_documents/mod.rs
@@ -30,7 +30,7 @@ use crate::update::{
     self, Facets, IndexerConfig, UpdateIndexingStep, WordPrefixDocids,
     WordPrefixPairProximityDocids, WordPrefixPositionDocids, WordsPrefixesFst,
 };
-use crate::{Index, Result, RoaringBitmapCodec};
+use crate::{Index, Result, RoaringBitmapCodec, UserError};
 
 static MERGED_DATABASE_COUNT: usize = 7;
 static PREFIX_DATABASE_COUNT: usize = 5;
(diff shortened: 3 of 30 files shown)


Kerollmops:
Could you fix the benchmarks, please?
https://github.com/meilisearch/milli/runs/5676613548?check_suite_focus=true#step:8:248

milli/src/update/index_documents/mod.rs
```diff
@@ -237,26 +251,36 @@ where
...
-                    searchable_fields,
-                    faceted_fields,
-                    primary_key_id,
-                    geo_field_id,
-                    stop_words,
-                    self.indexer_config.max_positions_per_attributes,
-                )
-            });
+            let result = original_chunk_iter
+                .and_then(|original_chunck_iter| Ok((original_chunck_iter, flattened_chunk_iter?)))
```
Kerollmops, line 268:
```suggestion
                .and_then(|original_chunk_iter| Ok((original_chunk_iter, flattened_chunk_iter?)))
```

milli/src/update/index_documents/transform.rs
```diff
@@ -96,22 +105,36 @@ impl<'a, 'i> Transform<'a, 'i> {
         };
 
         // We initialize the sorter with the user indexing settings.
-        let sorter = create_sorter(
+        let original_sorter = create_sorter(
             merge_function,
             indexer_settings.chunk_compression_type,
             indexer_settings.chunk_compression_level,
             indexer_settings.max_nb_chunks,
             indexer_settings.max_memory,
```
Kerollmops, line 113:
```suggestion
            indexer_settings.max_memory.map(|mem| mem / 2),
```

milli/src/update/index_documents/transform.rs
```diff
@@ -96,22 +105,36 @@ impl<'a, 'i> Transform<'a, 'i> {
...
         );
 
-        Transform {
+        // We initialize the sorter with the user indexing settings.
+        let flattened_sorter = create_sorter(
+            merge_function,
+            indexer_settings.chunk_compression_type,
+            indexer_settings.chunk_compression_level,
+            indexer_settings.max_nb_chunks,
+            indexer_settings.max_memory,
```
Kerollmops, line 122:
```suggestion
            indexer_settings.max_memory.map(|mem| mem / 2),
```

milli/src/update/index_documents/transform.rs
```diff
@@ -164,101 +194,215 @@ impl<'a, 'i> Transform<'a, 'i> {
...
+                    document,
+                    &addition_index,
+                    primary_key_id,
+                    &primary_key_name,
+                    &mut uuid_buffer,
+                    &mut field_buffer_cache,
+                    &mut external_id_buffer,
+                    self.autogenerate_docids,
+                )?
+            };
```
Kerollmops, line 229:
```suggestion
            }

            let external_id = update_primary_key(
                document,
                &addition_index,
                primary_key_id,
                &primary_key_name,
                &mut uuid_buffer,
                &mut field_buffer_cache,
                &mut external_id_buffer,
                self.autogenerate_docids,
            )?;
```

Kerollmops (changes requested):
I think that declaring these new databases isn't an issue, I find that this is a good way to do that. I haven't finished my review, I need to continue on the `transform.rs` file but I prefer publishing that than nothing.

milli/src/update/index_documents/transform.rs
```diff
@@ -164,101 +194,215 @@ impl<'a, 'i> Transform<'a, 'i> {
...
+                        match self
+                            .new_external_documents_ids_builder
+                            .entry(external_id.as_bytes().to_vec())
+                        {
+                            std::collections::hash_map::Entry::Occupied(entry) => {
+                                (*entry.get() as u32, false)
+                            }
+                            // if the document has never been encountered we give it a new docid
+                            // and push this new docid to the external documents ids builder
+                            std::collections::hash_map::Entry::Vacant(entry) => {
```
Kerollmops, line 260:
```suggestion
                            Entry::Occupied(entry) => {
                                (*entry.get() as u32, false)
                            }
                            // if the document has never been encountered we give it a new docid
                            // and push this new docid to the external documents ids builder
                            Entry::Vacant(entry) => {
```

milli/src/update/index_documents/transform.rs
```diff
@@ -164,101 +194,215 @@ impl<'a, 'i> Transform<'a, 'i> {
...
+    }
+
     /// Generate the `TransformOutput` based on the given sorter that can be generated from any
     /// format like CSV, JSON or JSON stream. This sorter must contain a key that is the document
     /// id for the user side and the value must be an obkv where keys are valid fields ids.
     pub(crate) fn output_from_sorter<F>(
         self,
         wtxn: &mut heed::RwTxn,
-        progress_callback: F,
+        _progress_callback: F,
```
Kerollmops, line 405: It could probably be an issue if we no more call this callback, don't you think? It is even easier to use it now that we have only have one loop, no?

milli/src/update/index_documents/transform.rs
```diff
@@ -268,140 +412,91 @@ impl<'a, 'i> Transform<'a, 'i> {
...
+                    // TODO: TAMO: remove unwrap
+                    .get(wtxn, &BEU32::new(u32_key))?
+                    .unwrap();
+
+                for (key, _) in obkv.iter() {
+                    let field =
+                        field_distribution.get_mut(self.fields_ids_map.name(key).unwrap()).unwrap();
+                    *field -= 1;
+                    if *field == 0 {
+                        field_distribution.remove(self.fields_ids_map.name(key).unwrap()).unwrap();
```
Kerollmops, line 446: Wow, there are too many `unwrap`s here. We must handle them!

milli/src/update/index_documents/transform.rs
```diff
@@ -268,140 +412,91 @@ impl<'a, 'i> Transform<'a, 'i> {
...
+
+        new_external_documents_ids_builder
+            .sort_unstable_by(|(left, _), (right, _)| left.cmp(&right));
+        let mut fst_new_external_documents_ids_builder = fst::MapBuilder::memory();
+        new_external_documents_ids_builder
+            .into_iter()
+            .try_for_each(|(key, value)| fst_new_external_documents_ids_builder.insert(key, value))
+            .unwrap();
+        let new_external_documents_ids = fst_new_external_documents_ids_builder.into_map();
+        external_documents_ids.insert_ids(&new_external_documents_ids).unwrap();
```
Kerollmops, line 488: Please handle those `unwrap`s with the `?` operator, here too.

milli/src/update/index_documents/transform.rs
```diff
@@ -445,13 +547,48 @@ impl<'a, 'i> Transform<'a, 'i> {
...
+                new_fields_ids_map.id(&key).unwrap_or(FieldId::MAX)
+            });
+
+            for (key, value) in flattened {
+                let fid =
+                    new_fields_ids_map.insert(&key).ok_or(UserError::AttributeLimitReached)?;
+                let value = serde_json::to_vec(&value).unwrap();
+                writer.insert(fid, &value)?;
+            }
+            flattened_writer.insert(docid.to_be_bytes(), &buffer).unwrap();
```
Kerollmops, line 582: Same: too many `unwrap`s.

Kerollmops (changes requested):
Could you please publish the benchmarks you ran recently here? And remove all of the `unwrap`s (or at least the problematic ones), this is probably something that you forgot to change 😃

milli/Cargo.toml
```diff
@@ -12,8 +12,10 @@ byteorder = "1.4.3"
 concat-arrays = "0.1.2"
 crossbeam-channel = "0.5.2"
 either = "1.6.1"
+fnv = "1.0.2"
 fst = "0.4.7"
 fxhash = "0.2.1"
```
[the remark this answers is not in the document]
irevoire: I got rid of fnv entirely and now only use `fxhash`

milli/src/lib.rs
```diff
@@ -183,6 +183,43 @@ pub fn lat_lng_to_xyz(coord: &[f64; 2]) -> [f64; 3] {
...
+    faceted_fields.into_iter().find(|facet| is_faceted_by(field, facet.as_ref())).is_some()
+}
+
+/// Returns `true` if the field match the facet.
+/// ```
+/// use milli::is_faceted_by;
+/// // -- the valid basics
+/// assert!(is_faceted_by("animaux", "animaux"));
+/// assert!(is_faceted_by("animaux.chien", "animaux"));
+/// assert!(is_faceted_by("animaux.chien.race.bouvier bernois.fourrure.couleur", "animaux"));
```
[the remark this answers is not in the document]
irevoire: Yes it can, but it won't work with something like that;
`animaux.chien.race."bouvier.bernois".fourrure.couleur`

milli/src/update/index_documents/transform.rs
irevoire: done :+1:

milli/src/update/index_documents/transform.rs
irevoire: done :+1:

milli/src/update/index_documents/transform.rs
irevoire: done :+1:

milli/src/update/index_documents/transform.rs
irevoire: I put the call back in the loop that compute the field distribution. Not sure it's the best but it's something.

irevoire: Maybe we should remove this callback entirely actually? Wdyt?

Kerollmops: Yeah, remove it if you don't use it, it would be better :)

milli/src/lib.rs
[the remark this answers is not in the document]
Kerollmops: Is it expected? Will it be fixed in a following PR?

Kerollmops (approved):
I approve this PR, it look good to me! Thanks for having added the `flatten-serde-json` crate in the repo along with the fuzzer! Now, we will see what people think about it in the RC0+ of Meilisearch 😃

bors merge

PR #460, opened 2022-03-04
Remove DS_Store files

brunoocasali:
Removes the annoying DS_Store files from macOS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 *.csv
 *.mmdb
 *.svg
+
+# macOS
+*.DS_Store


irevoire (commented):
I don’t know if that’s in the scope of the project 🤔 
Currently, our `.gitignore` is configured to ignore all the files generated by the project itself, and that’s all.

If we start to add the `.DStore` that is specific to macOS, then why not add the `Lost+Found` that we find on ubuntu, all the files that start with `.~` generated by emacs, the `.editor_config` and `.vscode` that visual studio code generate, `.idea` ... and when do we stop 😬 

Personally, I think milli should only handle his file correctly, and then if you have a specific config, it’s your job to configure your `.gitignore`, but not all the projects in the world to accommodate to your config.

Now that’s something that we never discussed, I think, so I would love to hear what the others think 👀 
@MarinPostma @Kerollmops @ManyTheFish

brunoocasali:
So, do you suggest something like this?

```bash
git config --global core.excludesfile ~/.gitignore

echo .DS_Store >> ~/.gitignore
```

You know, I never thought about this before! Reading more about the subject I found some interesting opinions about the subject:

- https://www.bmc.com/blogs/gitignore/
- https://julien.danjou.info/properly-managing-your-gitignore/

In this case, I'll close this PR, but I strongly advise creating a `CONTRIBUTING.md` and adding this info or should we use [meilisearch/CONTRIBUTING.md](https://github.com/meilisearch/meilisearch/blob/main/CONTRIBUTING.md)?

Thanks for the reply @irevoire changed my mind! 🎉

PR #461, opened 2022-03-04
Add a new error message when the `valid_fields` is empty

brunoocasali:
I've created a test case to handle the new error formatting behavior, but I'm not sure if:

- this is the right place to add the test?
- this is the best way to test this behavior?

And I'm not sure also regarding the `match` implementation, is this something required? Or maybe just an `if` statement is ok as well?

I left the two messages literally without "reusing the prefix" in the implementation because I think this could help the "searchability" of the error in the future.

# Pull Request

## What does this PR do?
Fixes https://github.com/meilisearch/meilisearch/issues/2140

## PR checklist
- [x] Does this PR fix an existing issue?
- [ ] Have you read the contributing guidelines?

### DIFF
--- a/milli/src/error.rs
+++ b/milli/src/error.rs
@@ -256,11 +256,20 @@ only composed of alphanumeric characters (a-z A-Z 0-9), hyphens (-) and undersco
             Self::InvalidSortableAttribute { field, valid_fields } => {
                 let valid_names =
                     valid_fields.iter().map(AsRef::as_ref).collect::<Vec<_>>().join(", ");
-                write!(
-                    f,
-                    "Attribute `{}` is not sortable. Available sortable attributes are: `{}`.",
-                    field, valid_names
-                )
+
+                    if valid_names.is_empty() {
+                        write!(
+                            f,
+                            "Attribute `{}` is not sortable. This index does not have configured sortable attributes.",
+                            field
+                        )
+                    } else {
+                        write!(
+                            f,
+                            "Attribute `{}` is not sortable. Available sortable attributes are: `{}`.",
+                            field, valid_names
+                        )
+                    }
             }
             Self::SortRankingRuleMissing => f.write_str(
                 "The sort ranking rule must be specified in the \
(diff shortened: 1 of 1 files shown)


milli/src/error.rs
```diff
@@ -256,11 +256,19 @@ only composed of alphanumeric characters (a-z A-Z 0-9), hyphens (-) and undersco
...
                 let valid_names =
                     valid_fields.iter().map(AsRef::as_ref).collect::<Vec<_>>().join(", ");
-                write!(
-                    f,
-                    "Attribute `{}` is not sortable. Available sortable attributes are: `{}`.",
-                    field, valid_names
-                )
+
+                    match valid_names.is_empty() {
+                        true => write!(
```
irevoire, line 261: Instead of doing a `match boolean { true => , false => }` you should do a `if boolean { } else { }` 👍

irevoire (changes requested):
Great thanks, there is a few tests that need to be updated it look like

brunoocasali:
> Great thanks, there is a few tests that need to be updated it look like

Yeah, I made the change and didn't run the tests again 😊

milli/src/error.rs
brunoocasali: Can you explain to me a little, when should I opt to use pattern matching or conditionals?

Kerollmops: It is considered _bad style_ or _unidiomatic_ to match on booleans, maybe because if-else is _the_ keyword to use to match on this type, or maybe because it is not a good thing to directly write the `true` and `false` words.

brunoocasali: Cool, I agree @Kerollmops! 

Btw I was looking for a way to match for `SomeValueInBtreeSet` and `None`, `_`, but I didn't find something like that, that could work with `BTreeSet`.

irevoire: Oh I think you can do something like that;
```rust
if let Some(value) = set.get(key) {
    // do something
}

Comment (Kerollmops) [reply milli/src/error.rs]: I am not sure that I understand what you wanted to do, could you explain in more detail? Or maybe provide a pseudo code?

Comment (brunoocasali) [reply milli/src/error.rs]:
```rust
match valid_names {
  HasValues => write!(...),
  EmptySet => write!(...),
}
```
Something like this (copied from meilisearch-rust):

```rust
   pub fn get_uid(&self) -> u64 {
        match self {
            Self::Enqueued { content } | Self::Processing { content } => *content.as_ref(),
            Self::Failed { content } => *content.as_ref(),
            Self::Succeeded { content } => *content.as_ref(),
        }
    }
```

Kerollmops: Ho ok, so something like this should have worked:
```rust
// You must match on the empty set first to make sure it is not caught by the unrestricted match.
match valid_names {
  [] => write!(...),
  values => write!(...),
}

// Another way to express this could be that way, but it is much more complex for no real gain.
match valid_names {
  [_, ..] @ values => write!(...),
  empty => write!(...),
}
```

You can read about [the @ _bindings_ in the official Rust book](https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html?highlight=match#-bindings).

curquiza:
Well done Brubru!!!

Bors merge

milli/src/error.rs
brunoocasali: Awesome, thanks for sharing that!

PR #462, opened 2022-03-07
cli improvements

MarinPostma:
a few improvements:
- use bufreader to load documents, so the loading of the document doesn't appear on flamegraphs
- set default db path to current directory so the `-i` flag can be omitted.


Kerollmops (approved):
Good for me, thank you!
bors merge

PR #472, opened 2022-03-16
Remove useless variables in proximity

ManyTheFish:
Was passing by plane sweep algorithm to find some inspiration, and I discover that we have useless variables that were not detected because of the recursive function.


Kerollmops (approved):
Wow! Indeed, it cleans the code a lot! thank you!
bors merge

PR #476, opened 2022-03-21
Rollback meilisearch-tokenizer version

irevoire:
Lindera often fails to download some data from google drive we can’t compile consistently meilisearch / milli.
We can’t bump to the latest version (that moved out of google drive) either because lindera uses reqwest with openssl with no way of configuring it our benchmarks were not able to run. The latter issue should be fixed by https://github.com/lindera-morphology/lindera/pull/164.


Kerollmops (approved):
Looks good to me, merging!
bors merge

PR #480, opened 2022-03-29
Increase benchmarks (push) CI timeout

Kerollmops:
This PR fixes the fact that the benchmarks CI on push were [canceled by GitHub](https://github.com/meilisearch/milli/actions/runs/2028844132) because they reached the default timeout of 6h. This PR changes the timeout to 72h, the same setting as the manually triggered benchmark one.


irevoire (approved):
Ah! Good catch that's what we had no benchmarks even after fixing lindera!

Kerollmops:
Sorry, I had to cancel the Ci to let #481 run correctly, I need to rerun the workflow when #481 is done!

Kerollmops:
bors merge

PR #487, opened 2022-04-07
Update version (v0.26.0)

curquiza:
breaking because of #458


Kerollmops (approved):
So now that #458 is merged we can merge this bump.
bors merge

PR #505, opened 2022-04-21
normalize exact words

MarinPostma:
Normalize the exact words, as specified in the specification.

### DIFF
--- a/milli/src/update/settings.rs
+++ b/milli/src/update/settings.rs
@@ -1461,4 +1461,22 @@ mod tests {
         builder.set_min_word_len_two_typos(7);
         assert!(builder.execute(|_| ()).is_err());
     }
+
+    #[test]
+    fn update_exact_words_normalization() {
+        let index = TempIndex::new();
+        let config = IndexerConfig::default();
+
+        // Set the genres setting
+        let mut txn = index.write_txn().unwrap();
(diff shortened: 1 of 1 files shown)


milli/src/update/settings.rs
```diff
@@ -580,6 +580,23 @@ impl<'a, 't, 'u, 'i> Settings<'a, 't, 'u, 'i> {
     fn update_exact_words(&mut self) -> Result<()> {
         match self.exact_words {
             Setting::Set(ref mut words) => {
+                fn normalize(analyzer: &Analyzer<&[u8]>, text: &str) -> String {
+                    analyzer.analyze(text).tokens().map(|token| token.text().to_string()).collect()
```
Kerollmops, line 584: Ok, so as I understand this function you are using the tokenizer to split and normalize the word for you, concatenating all of the normalized words together if there is more than one word. Is it expected?

Kerollmops (changes requested):
Could you fix the commit messages, please?

milli/src/update/settings.rs
MarinPostma: yes, for now @gmourier think it's better to ignore this edge case

Kerollmops (approved):
So if @gmourier is ok with that, then I approve those changes!

curquiza:
bors merge

PR #525, opened 2022-05-02
Simplify the error creation with thiserror

irevoire:
I introduced [`thiserror`](https://docs.rs/thiserror/latest/thiserror/) to implements all the `Display` trait and most of the `impl From<xxx> for yyy` in way less lines.
And then I introduced a cute macro to implements the `impl<X, Y, Z> From<X> for Z where Y: From<X>, Z: From<X>` more easily.


Kerollmops (approved):
Thank you @irevoire, it is much less cumbersome to introduce a new error now?
Thank you very much!

irevoire:
yes it should be easier :grin: 
bors merge

PR #531, opened 2022-05-16
fix the mixed dataset geosearch indexing bug

irevoire:
port #529 to main

### DIFF
--- a/milli/src/update/index_documents/extract/extract_geo_points.rs
+++ b/milli/src/update/index_documents/extract/extract_geo_points.rs
@@ -35,26 +35,28 @@ pub fn extract_geo_points<R: io::Read + io::Seek>(
         };
 
         // first we get the two fields
-        let lat = obkv
-            .get(lat_fid)
-            .ok_or_else(|| GeoError::MissingLatitude { document_id: primary_key() })?;
-        let lng = obkv
-            .get(lng_fid)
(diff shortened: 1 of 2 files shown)


milli/src/update/index_documents/extract/extract_geo_points.rs
```diff
@@ -35,26 +35,28 @@ pub fn extract_geo_points<R: io::Read + io::Seek>(
...
 
-        let bytes: [u8; 16] = concat_arrays![lat.to_ne_bytes(), lng.to_ne_bytes()];
-        writer.insert(docid_bytes, bytes)?;
+            let bytes: [u8; 16] = concat_arrays![lat.to_ne_bytes(), lng.to_ne_bytes()];
+            writer.insert(docid_bytes, bytes)?;
+        } else if lat.is_none() && lng.is_some() {
+            return Err(GeoError::MissingLatitude { document_id: primary_key() })?;
+        } else if lat.is_some() && lng.is_none() {
+            return Err(GeoError::MissingLongitude { document_id: primary_key() })?;
+        }
```
Kerollmops, line 59: Wouldn't this work? Or should we add more parentheses?

```suggestion
        } else if lat.is_none() && lng.is_some() || lat.is_some() && lng.is_none() {
            return Err(GeoError::MissingLatitude { document_id: primary_key() })?;
        }
```

Kerollmops (changes requested):

PR #535, opened 2022-05-18
Reintroduce the max values by facet limit

Kerollmops:
This PR reintroduces the max values by facet limit this is related to https://github.com/meilisearch/meilisearch/issues/2349.

~I would like some help in deciding on whether I keep the default 100 max values in milli and set up the `FacetDistribution` settings in Meilisearch to use 1000 as the new value, I expose the `max_values_by_facet` for this purpose.~

I changed the default value to 1000 and the max to 10000, thank you @ManyTheFish for the help!

### DIFF
--- a/milli/src/search/facet/facet_distribution.rs
+++ b/milli/src/search/facet/facet_distribution.rs
@@ -1,6 +1,6 @@
 use std::collections::{BTreeMap, HashSet};
 use std::ops::Bound::Unbounded;
-use std::{fmt, mem};
+use std::{cmp, fmt, mem};
 
 use heed::types::ByteSlice;
 use roaring::RoaringBitmap;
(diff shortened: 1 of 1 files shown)


milli/src/search/facet/facet_distribution.rs
```diff
@@ -52,6 +72,7 @@ impl<'a> FacetDistribution<'a> {
             FacetType::Number => {
                 let mut key_buffer: Vec<_> = field_id.to_be_bytes().iter().copied().collect();
 
+                let distribution_prelength = distribution.len();
```
ManyTheFish, line 75: In my point of view, it is only possible to have a non-empty `distribution` on String database, but this variable is used on Number database. 🤔
Is it a guard?

ManyTheFish (approved):
bors merge

PR #548, opened 2022-06-08
Setup the new limits on the number of facet values to return

Kerollmops:
This PR implements the early draft of the new spec (waiting for it) specifying how the new facet limit feature should work and which limit we apply to the number of facet values to return by facet.


ManyTheFish (approved):
Nice!

Bors merge

ManyTheFish (approved):
bors merge

PR #552, opened 2022-06-09
Fix escaped quotes in filter

irevoire:
Will fix https://github.com/meilisearch/meilisearch/issues/2380

The issue was that in the evaluation of the filter, I was using the deref implementation instead of calling the `value` method of my token.

To avoid the problem happening again, I removed the deref implementation; now, you need to either call the `lexeme` or the `value` methods but can't rely on a « default » implementation to get a string out of a token.

### DIFF
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "benchmarks"
-version = "0.29.2"
+version = "0.29.3"
 edition = "2018"
 publish = false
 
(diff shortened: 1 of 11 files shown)


curquiza:
Can you please update the version of milli so that we can release it immediately? 😇

Kerollmops (approved):
Nice to see this bug fixed! Thank you!
bors merge

Kerollmops:
bors merge-

curquiza:
bors merge-

PR #554, opened 2022-06-13
Ensure that Index methods are not bypassed by Meilisearch

ManyTheFish:
#### tests: (skip in changelog)
- [x] placeholder search shouldn’t return soft deleted
- [x] search shouldn’t return soft deleted
- [x] filtered placeholder search shouldn’t return soft deleted
- [x] geo-filtered placeholder search shouldn’t return soft deleted
- [x] documents list/get shouldn’t return soft deleted
- [x] stats shouldn’t count soft deleted

#### other: (API breaking)
- [x] ensure that Index methods are not bypassed by Meilisearch

Poke @irevoire, we may merge this into your branch.

### DIFF
--- a/infos/src/main.rs
+++ b/infos/src/main.rs
@@ -371,11 +371,9 @@ fn biggest_value_sizes(index: &Index, rtxn: &heed::RoTxn, limit: usize) -> anyho
     use std::cmp::Reverse;
     use std::collections::BinaryHeap;
 
-    use heed::types::{ByteSlice, Str};
+    use heed::types::ByteSlice;
 
     let Index {
-        env: _env,
-        main,
         word_docids,
         word_prefix_docids,
         docid_word_positions,
@@ -390,7 +388,7 @@ fn biggest_value_sizes(index: &Index, rtxn: &heed::RoTxn, limit: usize) -> anyho
         exact_word_prefix_docids,
         field_id_docid_facet_f64s: _,
         field_id_docid_facet_strings: _,
-        documents,
+        ..
     } = index;
 
     let main_name = "main";
@@ -425,11 +423,10 @@ fn biggest_value_sizes(index: &Index, rtxn: &heed::RoTxn, limit: usize) -> anyho
             heap.pop();
         }
 
-        if let Some(documents_ids) = main.get::<_, Str, ByteSlice>(rtxn, "documents-ids")? {
-            heap.push(Reverse((documents_ids.len(), format!("documents-ids"), main_name)));
-            if heap.len() > limit {
-                heap.pop();
-            }
+        let documents_ids = index.documents_ids(rtxn)?;
+        heap.push(Reverse((documents_ids.len() as usize, format!("documents-ids"), main_name)));
+        if heap.len() > limit {
+            heap.pop();
         }
 
         for result in word_docids.remap_data_type::<ByteSlice>().iter(rtxn)? {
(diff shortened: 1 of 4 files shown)


infos/src/main.rs
```diff
@@ -425,11 +423,10 @@ fn biggest_value_sizes(index: &Index, rtxn: &heed::RoTxn, limit: usize) -> anyho
             heap.pop();
         }
 
-        if let Some(documents_ids) = main.get::<_, Str, ByteSlice>(rtxn, "documents-ids")? {
-            heap.push(Reverse((documents_ids.len(), format!("documents-ids"), main_name)));
-            if heap.len() > limit {
-                heap.pop();
-            }
+        let documents_ids = index.documents_ids(rtxn)?;
+        heap.push(Reverse((documents_ids.len() as usize, format!("documents-ids"), main_name)));
```
irevoire, line 427:
```suggestion
        heap.push(Reverse((documents_ids.len() as usize, "documents-ids".to_string(), main_name)));
```

Since we don't format anything it's better to do a `to_string` I guess?

curquiza:
I removed the `skip-changelogs` label: since it's API breaking I definitely want this PR to be part of the changelogs 😄

irevoire (changes requested):

irevoire:
@curquiza this PR only adds tests to the current behaviour of milli, it's my PR that will break the DB :grin:

curquiza:
> @curquiza this PR only adds tests to the current behaviour of milli, it's my PR that will break the DB 😁

Will your PR be merged into this one (i.e. the `enhance-tests-for-soft-deletetion` branch)?
If yes -> let the `DB breaking` and remove the `skip-changelogs`
If no -> remove the `DB breaking` and put the `skip-changelogs` (if you want, it's ok if we let it in the changelogs too)

irevoire:
I would prefer merging this branch as-is into main since it doesn't break anything, and then I'll merge my branch later on main if @ManyTheFish is ok with that :+1:

milli/src/update/delete_documents.rs
```diff
@@ -768,82 +896,166 @@ mod tests {
...
+            {"id":"11","city":"Brussels",         "_geo": { "lat": 50.8466 as f32, "lng": 4.3370 as f32 } },
+            {"id":"12","city":"Charleroi",        "_geo": { "lat": 50.4095 as f32, "lng": 4.4347 as f32 } },
+            {"id":"13","city":"Mons",             "_geo": { "lat": 50.4502 as f32, "lng": 3.9623 as f32 } },
+            {"id":"14","city":"Valenciennes",     "_geo": { "lat": 50.3518 as f32, "lng": 3.5326 as f32 } },
+            {"id":"15","city":"Arras",            "_geo": { "lat": 50.2844 as f32, "lng": 2.7637 as f32 } },
+            {"id":"16","city":"Cambrai",          "_geo": { "lat": 50.1793 as f32, "lng": 3.2189 as f32 } },
+            {"id":"17","city":"Bapaume",          "_geo": { "lat": 50.1112 as f32, "lng": 2.8547 as f32 } },
+            {"id":"18","city":"Amiens",           "_geo": { "lat": 49.9314 as f32, "lng": 2.2710 as f32 } },
+            {"id":"19","city":"Compiègne",        "_geo": { "lat": 49.4449 as f32, "lng": 2.7913 as f32 } },
+            {"id":"20","city":"Paris",            "_geo": { "lat": 48.9021 as f32, "lng": 2.3708 as f32 } }
```
irevoire, line 918: Also, why are we specifying `as f32` everywhere? Can't we let rust infer whatever he wants?

irevoire (approved):
bors merge

PR #555, opened 2022-06-14
Improve the auto-batching error reporting

### LINKED ISSUE #546: Implement a better error reporting when using auto-batching
We discussed the issue with the auto-batching error reporting in https://github.com/meilisearch/meilisearch/issues/2478.

After discussing different ways to implement the above described expected behavior with @irevoire, @MarinPostma, and @gmourier, we continued on **how** we will implement this in the engine.

We decided to go on a pre-validation system, that would lay just before we insert the document into the transform system (L132). This way the validation system will be able to detect the possibly miss-formed documents (invalid primary key, __geo_ field) and return better error messages to the user.

https://github.com/meilisearch/milli/blob/05ae6dbfa40935414125d14f8d106edc87d74806/milli/src/update/index_documents/mod.rs#L118-L143

We also decided to change the function signature to take the ownership of the `IndexDocuments` struct and return something like an `internal::Result<(Self, user::Result<u64>)>`, where the `internal::Result` can throw a non-recoverable error and the `user::Result` can throw an error to the user, this user error will be visible in the tasks lists.

Kerollmops:
This PR fixes #546 by pre-checking the validity of the documents in the `IndexDocuments::add_documents` method before inserting them into the `Transform` system.

One of the first steps I want to do is to simplify the `DocumentsBatchBuilder/Reader` structs to use grenad instead of a custom serialization format because I wasn't ready to maintain it. To be able to pre-validate the documents it was mandatory to be able to read two times the same list of documents, grenad was already fulfilling this requirement, the customs system required a by-hand `seek` to zero. After moving to grenad we can see that the complexity and amount of code dropped.

Note that we need to store some metadata (the `DocumentsBatchIndex` struct) in this documents list, grenad doesn't support this feature, I, therefore, used a simple trick to store this metadata. I stored it as the absolute last key of the grenad file. The key isn't even in the range of the auto-incrementing id generated for each appended document, when this id is an `u32`, the metadata id is much bigger, it is an `u64::MAX` so 8 bytes and not 4, therefore, no conflict can happen.


Kerollmops:
I just ran a benchmark to check the time we take to index now. It should be slower as we check more things but #561 should improve that.

https://github.com/meilisearch/milli/actions/runs/2535467608

PR #556, opened 2022-06-14
Add EXISTS filter

loiclec:
## What does this PR do?

Fixes issue [#2484](https://github.com/meilisearch/meilisearch/issues/2484) in the meilisearch repo.

It creates a `field EXISTS` filter which selects all documents containing the `field` key. 
For example, with the following documents:
```json
[{
	"id": 0,
	"colour": []
},
{
	"id": 1,
	"colour": ["blue", "green"]
},
{
	"id": 2,
	"colour": 145238
},
{
	"id": 3,
	"colour": null
},
{
	"id": 4,
	"colour": {
		"green": []
	}
},
{
	"id": 5,
	"colour": {}
},
{
	"id": 6
}]
```
Then the filter `colour EXISTS` selects the ids `[0, 1, 2, 3, 4, 5]`. The filter `colour NOT EXISTS` selects `[6]`.

## Details
There is a new database named `facet-id-exists-docids`. Its keys are field ids and its values are bitmaps of all the document ids where the corresponding field exists.

To create this database, the indexing part of milli had to be adapted. The implementation there is basically copy/pasted from the code handling the `facet-id-f64-docids` database, with appropriate modifications in place.

There was an issue involving the flattening of documents during (re)indexing. Previously, the following JSON:
```json
{
    "id": 0,
    "colour": [],
    "size": {}
}
```
would be flattened to:
```json
{
    "id": 0
}
```
prior to being given to the extraction pipeline.

This transformation would lose the information that is needed to populate the `facet-id-exists-docids` database. Therefore, I have also changed the implementation of the `flatten-serde-json` crate. Now, as it traverses the Json, it keeps track of which key was encountered. Then, at the end, if a previously encountered key is not present in the flattened object, it adds that key to the object with an empty array as value. For example:
```json
{
    "id": 0,
    "colour": {
        "green": [],
        "blue": 1
    },
    "size": {}
} 
```
becomes
```json
{
    "id": 0,
    "colour": [],
    "colour.green": [],
    "colour.blue": 1,
    "size": []
} 
```


milli/src/index.rs
```diff
@@ -150,6 +154,9 @@ impl Index {
         let word_prefix_position_docids = env.create_database(Some(WORD_PREFIX_POSITION_DOCIDS))?;
         let facet_id_f64_docids = env.create_database(Some(FACET_ID_F64_DOCIDS))?;
         let facet_id_string_docids = env.create_database(Some(FACET_ID_STRING_DOCIDS))?;
+        let facet_id_exists_docids: Database<FieldIdCodec, CboRoaringBitmapCodec> =
+            env.create_database(Some(FACET_ID_EXISTS_DOCIDS))?;
```
irevoire, line 158:
```suggestion
        let facet_id_exists_docids = env.create_database(Some(FACET_ID_EXISTS_DOCIDS))?;
```

Do we really need to specify the type here?

milli/src/update/index_documents/extract/extract_facet_exists_docids.rs
```diff
@@ -0,0 +1,42 @@
...
+        max_memory,
+    );
+
+    let mut cursor = docid_fid_facet_number.into_cursor()?;
+    while let Some((key_bytes, _)) = cursor.move_on_next()? {
+        let (field_id, document_id) = FieldIdDocIdCodec::bytes_decode(key_bytes).unwrap();
+
+        let key_bytes = FieldIdCodec::bytes_encode(&field_id).unwrap();
+
+        facet_exists_docids_sorter.insert(key_bytes, document_id.to_ne_bytes())?;
```
irevoire, line 38:
```suggestion
    while let Some((key_bytes, _)) = cursor.move_on_next()? {
        let (field_id, document_id) = FieldIdDocIdCodec::bytes_decode(key_bytes).unwrap();
        let key_bytes = FieldIdCodec::bytes_encode(&field_id).unwrap();
        facet_exists_docids_sorter.insert(key_bytes, document_id.to_ne_bytes())?;
```

irevoire (changes requested):

milli/src/heed_codec/facet/mod.rs
```diff
@@ -25,3 +25,44 @@ pub fn try_split_at(slice: &[u8], mid: usize) -> Option<(&[u8], &[u8])> {
         None
     }
 }
+
+use std::borrow::Cow;
+use std::convert::TryInto;
+
+use crate::{try_split_array_at, DocumentId, FieldId};
+
+pub struct FieldIdCodec;
```
ManyTheFish, line 34: I think this implementation is not needed, you can use pre-implemented [Heed types](https://docs.rs/heed-types/latest/heed_types/) to encode primitive types. 😄

milli/src/update/index_documents/extract/extract_facet_exists_docids.rs
ManyTheFish: It could be worth initializing a reusable buffer before the loop and then encoding and decoding in this buffer. 🤔

milli/src/update/index_documents/extract/extract_facet_exists_docids.rs
```diff
@@ -0,0 +1,42 @@
...
+use crate::Result;
+
+/// Extracts the documents ids where this field appears.
+///
+/// Returns a grenad reader whose key is the field id encoded
+/// with `FieldIdCodec` and the value is a document_id (u32)
+/// encoded as native-endian bytes.
+#[logging_timer::time]
+pub fn extract_facet_exists_docids<R: io::Read + io::Seek>(
+    docid_fid_facet_number: grenad::Reader<R>,
```
ManyTheFish, line 19:
```suggestion
    docid_fid_facet_exists: grenad::Reader<R>,
```

milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
```diff
@@ -46,15 +54,19 @@ pub fn extract_fid_docid_facet_values<R: io::Read + io::Seek>(
...
-                let (numbers, strings) = extract_facet_values(&value);
-
                 key_buffer.clear();
 
+                // here, we know already that the document must be added to the “field id exists” database
                 // prefix key with the field_id and the document_id
+
                 key_buffer.extend_from_slice(&field_id.to_be_bytes());
                 key_buffer.extend_from_slice(&docid_bytes);
+                fid_docid_facet_exists_sorter.insert(&key_buffer, ().as_bytes())?;
```
ManyTheFish, line 64: I think we should benchmark it, but, shouldn't it be faster if we directly push the `field_id` as the key and the `docid` as the value replacing the `keep_first` merger with a `merge_cbo_roaring_bitmaps`?

milli/src/update/index_documents/mod.rs
```diff
@@ -1929,4 +1929,153 @@ mod tests {
...
+
+        wtxn.commit().unwrap();
+
+        let mut wtxn = index.write_txn().unwrap();
+        let mut builder = update::Settings::new(&mut wtxn, &index, &config);
+
+        let faceted_fields = hashset!(S("colour"));
+        builder.set_filterable_fields(faceted_fields);
+        builder.execute(|_| ()).unwrap();
+        wtxn.commit().unwrap();
```
ManyTheFish, line 1989: it is always better to put Settings before documents. 😛

milli/src/update/index_documents/mod.rs
```diff
@@ -1929,4 +1929,153 @@ mod tests {
...
+        let colour_id = index.fields_ids_map(&rtxn).unwrap().id("colour").unwrap();
+        let colour_green_id = index.fields_ids_map(&rtxn).unwrap().id("colour.green").unwrap();
+
+        let bitmap_colour = index.facet_id_exists_docids.get(&rtxn, &colour_id).unwrap().unwrap();
+        assert_eq!(bitmap_colour.into_iter().collect::<Vec<_>>(), vec![0, 1, 2, 3, 4, 6]);
+
+        let bitmap_colour_green =
+            index.facet_id_exists_docids.get(&rtxn, &colour_green_id).unwrap().unwrap();
+        assert_eq!(bitmap_colour_green.into_iter().collect::<Vec<_>>(), vec![6]);
+    }
```
ManyTheFish, line 2080: could you merge both tests in one, please ?

ManyTheFish (changes requested):

milli/src/heed_codec/facet/mod.rs
```diff
@@ -25,3 +25,44 @@ pub fn try_split_at(slice: &[u8], mid: usize) -> Option<(&[u8], &[u8])> {
...
+    fn bytes_decode(bytes: &'a [u8]) -> Option<Self::DItem> {
+        let (field_id_bytes, bytes) = try_split_array_at(bytes)?;
+        let field_id = u16::from_be_bytes(field_id_bytes);
+
+        let document_id_bytes = bytes[..4].try_into().ok()?;
+        let document_id = u32::from_be_bytes(document_id_bytes);
+
+        Some((field_id, document_id))
+    }
+}
```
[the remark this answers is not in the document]
Kerollmops: Hum... I don't remember we should check the codecs that we use in the `Index` database.

milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
loiclec: I had some problems when I tried to do that at first because, but I tried again this morning and it seems to work :-)

milli/src/heed_codec/facet/mod.rs
loiclec: The Heed types don't implement `BytesEncode`/`BytesDecode`, but I am now using them in the implementation of `FieldCodecId` for clarity :)

milli/src/heed_codec/facet/mod.rs
```diff
@@ -25,3 +25,44 @@ pub fn try_split_at(slice: &[u8], mid: usize) -> Option<(&[u8], &[u8])> {
...
+
+impl<'a> heed::BytesEncode<'a> for FieldIdCodec {
+    type EItem = FieldId;
+
+    fn bytes_encode(field_id: &Self::EItem) -> Option<Cow<[u8]>> {
+        Some(Cow::Owned(field_id.to_be_bytes().to_vec()))
+    }
+}
+
+pub struct FieldIdDocIdCodec;
```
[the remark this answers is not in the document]
loiclec: It wasn't needed. I am now implementing it anyway, for consistency's sake :)

irevoire (approved):
looks good to me thank you :+1:

milli/src/heed_codec/facet/field_id_codec.rs
```diff
@@ -0,0 +1,28 @@
...
+
+impl<'a> heed::BytesEncode<'a> for FieldIdCodec {
+    type EItem = FieldId;
+
+    fn bytes_encode(field_id: &Self::EItem) -> Option<Cow<[u8]>> {
+        let field_id = BEU16::new(*field_id);
+        let bytes = field_id.as_bytes();
+        Some(Cow::Owned(bytes.to_vec()))
+    }
+}
```
Kerollmops, line 28: Why not directly declare a `type FieldIdCodec = BEU16` here, instead?

Kerollmops (changes requested):

milli/src/heed_codec/facet/field_id_codec.rs
loiclec: When I looked into it, I found that `BEU16` didn't implement `BytesEncode`/`Decode`.

Kerollmops: You should probably [use `OwnedType<BEU16>` instead sorry](https://docs.rs/heed-types/0.8.0/heed_types/struct.OwnedType.html).

milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
```diff
@@ -79,7 +93,8 @@ pub fn extract_fid_docid_facet_values<R: io::Read + io::Seek>(
 
     Ok((
         sorter_into_reader(fid_docid_facet_numbers_sorter, indexer.clone())?,
-        sorter_into_reader(fid_docid_facet_strings_sorter, indexer)?,
+        sorter_into_reader(fid_docid_facet_strings_sorter, indexer.clone())?,
+        facet_exists_docids,
```
ManyTheFish, line 97: I'm a bit concerned by this btreemap return, I understand that its size is small, but, we are in a multi-threaded process where threads share and maximize the usage of memory. I feel like the memory used by this btreemap "leaks" out of the process.
I suggest finding a way to write the data on the disk before quitting the function.
I don't know if I'm clear, so don't hesitate to ask me to re-phrase it.

ManyTheFish (changes requested):

milli/src/update/index_documents/extract/extract_fid_docid_facet_values.rs
loiclec: You're right. I did think it was too small to matter at the beginning, but if we have 10 million documents and 100 faceted fields, it could get quite large! I will rework it :)

loiclec:
@ManyTheFish thanks for your review :) 

A big PR was merged into main since then and I had to make this PR up-to-date with the new main. Rebasing was quite difficult to do so instead I merged the previous filter/field-exists and main, resolved the conflicts, and pushed the result again under the same branch. I hope it's not too messy :(

filter-parser/fuzz/.gitignore
```diff
@@ -1,2 +1,3 @@
 /corpus/
 /artifacts/
+/target/
```
ManyTheFish, line 3:
```suggestion
/target/

```

ManyTheFish (dismissed):
Ouch! let's merge like this.
But! In this case of big merge conflicts, I prefer to squash all my commits before rebasing. Yes, I lose my history but I find it way clearer!

ManyTheFish (approved):
Hello @loiclec, sorry for the time!

You can merge it if everything is ok on your side.

loiclec:
Thanks everyone!
bors merge

PR #557, opened 2022-06-15
Fasten the document deletion

irevoire:
When a document deletion occurs, instead of deleting the document we mark it as deleted in the new “soft deleted” bitmap. It is then removed from the search and all the other endpoints.


milli/src/index.rs
```diff
@@ -249,6 +250,29 @@ impl Index {
         Ok(count.unwrap_or_default())
     }
 
+    /* deleted documents ids */
+
+    /// Writes the documents ids that corresponds to the user-ids-documents-ids FST.
```
Kerollmops, line 255: Whoops, wrong descriptions.

milli/src/index.rs
```diff
@@ -249,6 +250,29 @@ impl Index {
...
+        docids: &RoaringBitmap,
+    ) -> heed::Result<()> {
+        self.main.put::<_, Str, RoaringBitmapCodec>(
+            wtxn,
+            main_key::SOFT_DELETED_DOCUMENTS_IDS_KEY,
+            docids,
+        )
+    }
+
+    /// Returns the internal documents ids.
```
Kerollmops, line 268: Whoops, wrong description.

Kerollmops (changes requested):

milli/src/search/facet/filter.rs
```diff
@@ -266,11 +264,12 @@ impl<'a> Filter<'a> {
...
         rtxn: &heed::RoTxn,
         index: &Index,
-        numbers_db: heed::Database<FacetLevelValueF64Codec, CboRoaringBitmapCodec>,
-        strings_db: heed::Database<FacetStringLevelZeroCodec, FacetStringLevelZeroValueCodec>,
         field_id: FieldId,
         operator: &Condition<'a>,
     ) -> Result<RoaringBitmap> {
+        let numbers_db = index.facet_id_f64_docids;
+        let strings_db = index.facet_id_string_docids;
+
```
Kerollmops, line 272: This change is only here to make the code clearer, right?

milli/src/update/available_documents_ids.rs
```diff
@@ -8,11 +8,16 @@ pub struct AvailableDocumentsIds {
...
+        docids: &RoaringBitmap,
+        soft_deleted_docids: &RoaringBitmap,
+    ) -> AvailableDocumentsIds {
+        let used_docids = docids | soft_deleted_docids;
+
+        match used_docids.max() {
             Some(last_id) => {
                 let mut available = RoaringBitmap::from_iter(0..last_id);
-                available -= docids;
+                available -= used_docids;
```
Kerollmops, line 20: Could you do that in one line, please? Put the OR on the `used_docids` line?

milli/src/update/index_documents/mod.rs
```diff
@@ -719,10 +719,11 @@ mod tests {
         assert_eq!(count, 1);
 
         // Check that we get only one document from the database.
-        let docs = index.documents(&rtxn, Some(0)).unwrap();
+        // Since the document has been deleted and re-inserted, its internal docid has has been incremented to 1
```
Kerollmops, line 722:
```suggestion
        // Since the document has been deleted and re-inserted, its internal docid has been incremented to 1
```

milli/src/update/delete_documents.rs
```diff
@@ -17,15 +16,17 @@ use crate::heed_codec::facet::{
 use crate::heed_codec::CboRoaringBitmapCodec;
 use crate::index::{db_name, main_key};
 use crate::{
-    DocumentId, ExternalDocumentsIds, FieldId, Index, Result, RoaringBitmapCodec, SmallString32,
-    BEU32,
+    DocumentId, ExternalDocumentsIds, FieldId, FieldIdMapMissingEntry, Index, Result,
+    RoaringBitmapCodec, SmallString32, BEU32,
 };
 
+const DELETE_DOCUMENTS_THRESHOLD: u64 = 100_000;
```
Kerollmops, line 23: Could you add a description to this _const_ please? What is it used for?

milli/src/update/delete_documents.rs
```diff
@@ -93,6 +96,50 @@ impl<'t, 'u, 'i> DeleteDocuments<'t, 'u, 'i> {
...
+                    })?;
+                if let Entry::Occupied(mut entry) = field_distribution.entry(field_name.to_string())
+                {
+                    match entry.get().checked_sub(1) {
+                        Some(0) | None => entry.remove(),
+                        Some(count) => entry.insert(count),
+                    };
+                }
+            }
+        }
```
Kerollmops, line 122: I would prefer that we don't add this part of the code here, I would prefer that we store the docids corresponding to each field in the field distribution instead. This way when we retrieve the distribution of the fields we can simply remove the soft-deleted once. Could you change that and add a commit for the previously described behavior, please?

Kerollmops: Hum... in fact could you please open an issue about doing that later?

Kerollmops (changes requested):

milli/src/search/facet/filter.rs
irevoire: Yep, I noticed we were passing this by parameters everywhere when it was actually available directly from the `Index` without any function call (which has _similar_ performance).
And it’s easier to read without theses two parameters 😁

milli/src/update/available_documents_ids.rs
irevoire: Not sure I understood. Do you want something like that 🤔?
```rust
available -= docids | soft_deleted_docids;
```

If yes, did you miss the fact that I also use the `used_docids` in the match?

Kerollmops (approved):
I accept this PR and will merge once we are ready, which is approximately in 1 week.

Kerollmops (approved):
bors merge

ManyTheFish (changes requested):

milli/src/index.rs
```diff
@@ -929,9 +954,13 @@ impl Index {
         rtxn: &'t RoTxn,
         ids: impl IntoIterator<Item = DocumentId>,
     ) -> Result<Vec<(DocumentId, obkv::KvReaderU16<'t>)>> {
+        let soft_deleted_documents = self.soft_deleted_documents_ids(rtxn)?;
         let mut documents = Vec::new();
 
         for id in ids {
+            if soft_deleted_documents.contains(id) {
+                return Err(InternalError::AccessingSoftDeletedDocument { document_id: id })?;
```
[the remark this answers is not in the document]
Kerollmops: Yeah! I am d'accord with you!

ManyTheFish (approved):

milli/src/index.rs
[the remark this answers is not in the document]
ManyTheFish: @irevoire I agree with every error but not an internal one.

PR #558, opened 2022-06-16
Deletion benchmarks

ManyTheFish:
Add benchmarks on the deletion and start rethinking benchmark names.


irevoire (approved):
thanks for the factorization, the file was starting to get really large :+1:

ManyTheFish:
bors merge

PR #561, opened 2022-06-20
Enriched documents batch reader

Kerollmops:
~This PR is based on #555 and must be rebased on main after it has been merged to ease the review.~
This PR contains the work in #555 and can be merged on main as soon as reviewed and approved.

- [x] Create an `EnrichedDocumentsBatchReader` that contains the external documents id.
- [x] Extract the primary key name and make it accessible in the `EnrichedDocumentsBatchReader`.
- [x] Use the external id from the `EnrichedDocumentsBatchReader` in the `Transform::read_documents`.
- [x] Remove the `update_primary_key` from the _transform.rs_ file.
- [x] Really generate the auto-generated documents ids.
- [x] Insert the (auto-generated) document ids in the document while processing it in `Transform::read_documents`.


loiclec (commented):
It is a fairly big PR for me, so I did my best, but overall I think it looks good :) There are mostly two potential issues:

* should we verify that the floats inside the geo search fields are valid longitude/latitudes, for example by checking that they are not NaN, inf, etc.?
* I think there might be a bug in the code that finds the nested primary key

Apart from that, I had two minor comments:
1. we'll need to reintroduce the functionality provided by `serde_impl.rs`, as I don't think it will be acceptable, performance-wise, to deserialise the whole array of added documents at once
2. the “guessed” primary ID uses the whole `DocumentsBatchIndex`, but the Meilisearch documentation says that the guess should only use the first document. This is not a change in behaviour introduced by this PR though, so it's not blocking  as far as I am concerned

milli/src/update/index_documents/enrich.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+}
+
+pub fn validate_geo_from_json(id: &DocumentId, bytes: &[u8]) -> Result<StdResult<(), GeoError>> {
+    use GeoError::*;
+    let debug_id = || Value::from(id.debug());
+    match serde_json::from_slice(bytes).map_err(InternalError::SerdeJson)? {
+        Value::Object(mut object) => match (object.remove("lat"), object.remove("lng")) {
+            (Some(lat), Some(lng)) => {
+                match (extract_float_from_value(lat), extract_float_from_value(lng)) {
```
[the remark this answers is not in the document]
Kerollmops: That's a good idea indeed, thank you!

milli/src/update/index_documents/enrich.rs
```diff
@@ -0,0 +1,358 @@
...
+        }
+        nested @ PrimaryKey::Nested { .. } => {
+            let mut matching_documents_ids = Vec::new();
+            for (first_level_name, right) in nested.possible_level_names() {
+                if let Some(field_id) = documents_batch_index.id(first_level_name) {
+                    if let Some(value_bytes) = document.get(field_id) {
+                        let object = serde_json::from_slice(value_bytes)
+                            .map_err(InternalError::SerdeJson)?;
+                        fetch_matching_values(object, right, &mut matching_documents_ids);
+
```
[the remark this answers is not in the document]
Kerollmops: It should be fixed now, thank you very much for this deep bug finding ❤️

milli/src/documents/serde_impl.rs
```diff
@@ -1,134 +0,0 @@
-use std::collections::BTreeMap;
```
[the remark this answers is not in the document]
Kerollmops: I just published a new commit that introduces the `append_json_array` function, could you check it, please?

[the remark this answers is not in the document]
loiclec: Yes, I am pretty sure it's correct :)

loiclec:
Thank you for the changes! :) 

Regarding the nested primary keys again. Could you check that the function works correctly when we have the document:
```json
{ "a" : { "b" : { "c" :  1 }}}
```
and the primary keys `a.b` or `a.b.c.d`?
It is my understanding that both would select the value `Number(1)`.

loiclec:
It looks good to me! :)

loiclec (approved):
Thanks! I can't spot anything that's wrong, although I mostly looked at the transform file and (relatively) quickly read through the changes since the last review.

milli/src/update/index_documents/enrich.rs
```diff
@@ -0,0 +1,364 @@
...
+pub fn validate_document_id(document_id: &str) -> Option<&str> {
+    let document_id = document_id.trim();
+    if !document_id.is_empty()
+        && document_id.chars().all(|c| matches!(c, 'a'..='z' | 'A'..='Z' | '0'..='9' | '-' | '_'))
+    {
+        Some(document_id)
+    } else {
+        None
+    }
+}
```
irevoire, line 306: Was this already the behaviour?

If I send
```json
{ "id": "1", "pet": "doggo" }
{ "id": " 1", "pet": "catto" }
```
the ` 1` is going to replace the `1`?

[I thought spaces were invalid in document ids](https://docs.meilisearch.com/learn/core_concepts/primary_key.html#primary-field);
> The document id must be an integer or a string. If the id is a string, it can only contain alphanumeric characters (a-z, A-Z, 0-9), hyphens (-), and underscores (_).
[#](https://docs.meilisearch.com/learn/core_concepts/primary_key.html#setting-the-primary-key)

milli/src/update/index_documents/transform.rs
```diff
@@ -205,8 +191,7 @@ impl<'a, 'i> Transform<'a, 'i> {
             // it, transform it into a string and validate it, and then update it in the
             // document. If none is found, and we were told to generate missing document ids, then
             // we create the missing field, and update the new document.
-            let mut uuid_buffer = [0; uuid::fmt::Hyphenated::LENGTH];
-            let external_id = if primary_key_id_nested {
+            if primary_key_id_nested {
```
irevoire, line 194: I think we can get rid of this part too and delete the `flatten_from_field_mapping` function!
We'll flatten the document later with the function that works with the field id map :grin: 

Also, in case we need to keep this if, the comment above is outdated

irevoire (changes requested):
Overall it's a super cool PR!
I think we can simplify the transform even more with your enriched batch reader :grin:

milli/src/update/index_documents/transform.rs
Kerollmops: Hum... Thank you indeed, it maybe emerged from the rebase I did on main.

milli/src/update/index_documents/enrich.rs
Kerollmops: I didn't change the behavior, [it is the same as before](https://github.com/meilisearch/milli/blob/ebddfdb9a329e955066b15ffae17c3a4d496e9c9/milli/src/update/index_documents/transform.rs#L783-L789).

Kerollmops:
> I think we can simplify the transform even more with your enriched batch reader 😁

Thank you for your review 🙏, would it be possible for you to do that next week, as I am on vacation, please?

milli/src/update/index_documents/enrich.rs
irevoire: but it is wrong, right?

milli/src/update/index_documents/transform.rs
loiclec: I have removed this section and the code that supported it :)

milli/src/update/index_documents/enrich.rs
loiclec: @ManyTheFish I would like to have your opinion on this. Should we merge the PR but create an issue about it?

ManyTheFish: @irevoire is right, this seems to be a bug. However, as you said @loiclec, let's do it in another PR and create an issue.

This PR is already too long 😅

edit1: if someone took the time to add this trim, we should discuss the "why" before "fixing" it.

ManyTheFish (changes requested):

loiclec:
@ManyTheFish I committed your suggested change to add a comment. Was there anything else to do?

ManyTheFish (approved):

curquiza:
Checked with @loiclec, we can finally merge!!!

bors merge

PR #566, opened 2022-06-22
Introduce the copy_to_path method on the Index

Kerollmops:
Meilisearch needs this method to do snapshots.


irevoire (approved):
ah yes since the `env` is now private :unamused: 
thank you
bors merge

PR #568, opened 2022-06-27
Fix not equal filter when field contains both number and strings

GraDKh:
Related to https://github.com/meilisearch/meilisearch/issues/2516
Looks like the issue should be moved to this repo, but I'm not sure what the right procedure for it.


Kerollmops (approved):
Thank you very much, it indeed makes the filters valid now! However, I removed the _fixes_ keyword from your PR description as it is not fixed in Meilisearch, yet! 😃
bors merge

PR #572, opened 2022-07-04
Add reindexing benchmarks

irevoire:
With #557 coming, we should add benchmarks that measure our impact on the reindexing process.


ManyTheFish (dismissed):

irevoire:
@ManyTheFish I rebased on main

Kerollmops (approved):
Thank you very much! Looks good to me!
bors merge

PR #577, opened 2022-07-05
Fix deserialisation of NDJson documents in benchmarks

loiclec:
Previously, the first document in the NDJson file was read over and over again. So the `geo_point` benchmark was not working properly: it only indexed one document.


irevoire (approved):
Thanks @loiclec :dog2: 
bors merge

PR #580, opened 2022-07-06
Categorize the os error codes 5 and 28 as storage full user errors

Kerollmops:
This PR is related to https://github.com/meilisearch/meilisearch/issues/2255 and correctly categorizes the os error codes 5 and 28 as a `UserError::NoSpaceLeftOnDevice` error. We will display the message: _There is no more space left on the device. Consider increasing the size of the disk/partition_ when we encounter one of those two os error codes.

### DIFF
--- a/milli/src/error.rs
+++ b/milli/src/error.rs
@@ -9,6 +9,11 @@ use thiserror::Error;
 
 use crate::{CriterionError, DocumentId, FieldId, SortError};
 
+/// The raw os error code for the "No Space Left On Device" error.
+const STORAGE_FULL_ERROR_CODE: i32 = 28;
+/// The raw os error code for the "Input/Output error" error.
+const INPUT_OUTPUT_ERROR_CODE: i32 = 5;
+
 pub type Object = Map<String, Value>;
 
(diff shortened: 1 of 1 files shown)


irevoire (approved):
:+1:

curquiza:
@Kerollmops check it indeed returns the code error `no_space_left_on_device` https://github.com/meilisearch/specifications/blob/main/text/0061-error-format-and-definitions.md#no_space_left_on_device

ManyTheFish:
Hey @curquiza, the error code can't be returned by [Meilisearch directly but could come from milli](https://github.com/meilisearch/meilisearch/search?q=NoSpaceLeftOnDevice), but [milli never uses this error](https://github.com/meilisearch/milli/search?q=NoSpaceLeftOnDevice), so this error code is never returned.
However, there are some ways that an `Internal Error`  could be returned with a message containing something equivalent to `no space left on device`, we write in a file at least during the snapshot and dump process when writing the snapshot/dump, and the indexing process via grenad. Unfortunately, there is no specialized error variant equivalent to our `no_space_left_on_device` so far making it difficult to bind the error.

PR #583, opened 2022-07-06
Use BufReader to read datasets in benchmarks

loiclec:
## What does this PR do?
Ensure that the datasets used by the benchmarks are read efficiently by using a `BufReader`.

## Why?
Using a `BufReader` is more representative of how `meilisearch` works. It will also make performance comparisons between different branches of `milli` more  accurate.


ManyTheFish (approved):
Purfect! 

bors merge

PR #587, opened 2022-07-14
Word prefix pair proximity docids indexation refactor

loiclec:
# Pull Request

## What does this PR do?
Refactor the code of `WordPrefixPairProximityDocIds` to make it much faster, fix a bug, and add a unit test.

## Why is it faster?
Because we avoid using a sorter to insert the (`word1`, `prefix`, `proximity`) keys and their associated bitmaps, and thus we don't have to sort a potentially very big set of data. I have also added a couple of other optimisations: 

1. reusing allocations
2. using a prefix trie instead of an array of prefixes to get all the prefixes of a word
3. inserting directly into the database instead of putting the data in an intermediary grenad when possible. Also avoid checking for pre-existing values in the database when we know for certain that they do not exist. 

## What bug was fixed?
When reindexing, the `new_prefix_fst_words` prefixes may look like:
```
["ant",  "axo", "bor"]
```
which we group by first letter:
```
[["ant", "axo"], ["bor"]]
```

Later in the code, if we have the word2 "axolotl", we try to find which subarray of prefixes contains its prefixes. This check is done with `word2.starts_with(subarray_prefixes[0])`, but `"axolotl".starts_with("ant")` is false, and thus we wrongly think that there are no prefixes in `new_prefix_fst_words` that are prefixes of `axolotl`.

## StrStrU8Codec
I had to change the encoding of `StrStrU8Codec` to make the second string null-terminated as well. I don't think this should be a problem, but I may have missed some nuances about the impacts of this change.

## Requests when reviewing this PR
I have explained what the code does in the module documentation of `word_pair_proximity_prefix_docids`. It would be nice if someone could read it and give their opinion on whether it is a clear explanation or not. 

I also have a couple questions regarding the code itself:
- Should we clean up and factor out the `PrefixTrieNode` code to try and make broader use of it outside this module? For now, the prefixes undergo a few transformations: from FST, to array, to prefix trie. It seems like it could be simplified.
- I wrote a function called `write_into_lmdb_database_without_merging`. (1) Are we okay with such a function existing? (2) Should it be in `grenad_helpers` instead?

## Benchmark Results

We reduce the time it takes to index about 8% in most cases, but it varies between -3% and -20%. 

```
group                                                                     indexing_main_ce90fc62                  indexing_word-prefix-pair-proximity-docids-refactor_cbad2023
-----                                                                     ----------------------                  ------------------------------------------------------------
indexing/-geo-delete-facetedNumber-facetedGeo-searchable-                 1.00  1893.0±233.03µs        ? ?/sec    1.01  1921.2±260.79µs        ? ?/sec
indexing/-movies-delete-facetedString-facetedNumber-searchable-           1.05      9.4±3.51ms        ? ?/sec     1.00      9.0±2.14ms        ? ?/sec
indexing/-movies-delete-facetedString-facetedNumber-searchable-nested-    1.22    18.3±11.42ms        ? ?/sec     1.00     15.0±5.79ms        ? ?/sec
indexing/-songs-delete-facetedString-facetedNumber-searchable-            1.00     41.4±4.20ms        ? ?/sec     1.28    53.0±13.97ms        ? ?/sec
indexing/-wiki-delete-searchable-                                         1.00   285.6±18.12ms        ? ?/sec     1.03   293.1±16.09ms        ? ?/sec
indexing/Indexing geo_point                                               1.03      60.8±0.45s        ? ?/sec     1.00      58.8±0.68s        ? ?/sec
indexing/Indexing movies in three batches                                 1.14      16.5±0.30s        ? ?/sec     1.00      14.5±0.24s        ? ?/sec
indexing/Indexing movies with default settings                            1.11      13.7±0.07s        ? ?/sec     1.00      12.3±0.28s        ? ?/sec
indexing/Indexing nested movies with default settings                     1.10      10.6±0.11s        ? ?/sec     1.00       9.6±0.15s        ? ?/sec
indexing/Indexing nested movies without any facets                        1.11       9.4±0.15s        ? ?/sec     1.00       8.5±0.10s        ? ?/sec
indexing/Indexing songs in three batches with default settings            1.18      66.2±0.39s        ? ?/sec     1.00      56.0±0.67s        ? ?/sec
indexing/Indexing songs with default settings                             1.07      58.7±1.26s        ? ?/sec     1.00      54.7±1.71s        ? ?/sec
indexing/Indexing songs without any facets                                1.08      53.1±0.88s        ? ?/sec     1.00      49.3±1.43s        ? ?/sec
indexing/Indexing songs without faceted numbers                           1.08      57.7±1.33s        ? ?/sec     1.00      53.3±0.98s        ? ?/sec
indexing/Indexing wiki                                                    1.06   1051.1±21.46s        ? ?/sec     1.00    989.6±24.55s        ? ?/sec
indexing/Indexing wiki in three batches                                   1.20    1184.8±8.93s        ? ?/sec     1.00     989.7±7.06s        ? ?/sec
indexing/Reindexing geo_point                                             1.04      67.5±0.75s        ? ?/sec     1.00      64.9±0.32s        ? ?/sec
indexing/Reindexing movies with default settings                          1.12      13.9±0.17s        ? ?/sec     1.00      12.4±0.13s        ? ?/sec
indexing/Reindexing songs with default settings                           1.05      60.6±0.84s        ? ?/sec     1.00      57.5±0.99s        ? ?/sec
indexing/Reindexing wiki                                                  1.07   1725.0±17.92s        ? ?/sec     1.00    1611.4±9.90s        ? ?/sec
```


ManyTheFish (changes requested):
Nice work, I like the code structure.
However, This change is really huge, and, I'd like to have the feedback of someone else like kerollmops before merging this.

milli/src/update/word_prefix_pair_proximity_docids.rs
```diff
@@ -169,78 +373,549 @@ impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
...
+        search_start: &PrefixTrieNodeSearchStart,
+        mut do_fn: impl FnMut(&mut Vec<u8>),
+    ) {
+        let first_byte = word[0];
+        let mut cur_node = self;
+        buffer.push(first_byte);
+        if let Some((child_node, c)) =
+            cur_node.children[search_start.0..].iter().find(|(_, c)| *c >= first_byte)
+        {
+            if *c == first_byte {
```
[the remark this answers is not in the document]
loiclec: Good point. Part of the answer is that it is a leftover from a previous refactor which does not make as much sense anymore.

But still...
1. In pretty much all cases, `cur_node.children[search_start.0]` is the node we want to find. So a binary search would be a lot less efficient.
2. If we use `.find(|..| c == first_byte)` and there is no child for which the condition is true, then we will iterate over all children, which is also less efficient. Doing `c >= first_byte` allows us to stop searching early.

These kinds of changes are hard to benchmark. I wanted to strike a balance between something that is simple, fast in the common case, and never too slow. I think I'm in favour of replacing the inner loop by a binary search, but also to keep the first condition as it is.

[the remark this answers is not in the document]
ManyTheFish: Makes sense

milli/src/update/word_prefix_pair_proximity_docids.rs
```diff
@@ -169,78 +373,549 @@ impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
...
+        self.byte_vectors.push(data);
+    }
+    fn reclaim_mergeable_data_vector(&mut self, mut data: Vec<Cow<'static, [u8]>>) {
+        data.clear();
+        self.mergeable_data_vectors.push(data);
+    }
+}
+
+#[derive(Default, Debug)]
+struct PrefixTrieNode {
```
Kerollmops, line 625: Would it be possible to explain how this struct works, with a simple doc comment, please?

infos/src/main.rs
```diff
@@ -1138,6 +1138,7 @@ fn word_pair_proximities_docids(
     prefix.extend_from_slice(word1.as_bytes());
     prefix.push(0);
     prefix.extend_from_slice(word2.as_bytes());
+    prefix.push(0);
```
Kerollmops, line 1141: Are we sure that we want to add this byte at the end, this way we change the serialization format and deserializers must be all synchronized.

Kerollmops (changes requested):

milli/src/update/word_prefix_pair_proximity_docids.rs
```diff
@@ -1,16 +1,210 @@
...
+docids from the batch directly into the database (we would have a concurrent
+reader and writer). Therefore, when calling the algorithm on
+(`new_prefixes`, `word_pairs_db`), we insert the computed
+((`word`, `prefix`, `proximity`), `docids`) elements in an intermediary grenad
+Writer instead of the DB. At the end of the outer loop, we finally read from
+the grenad and insert its elements in the database.
+
+
+
+*/
```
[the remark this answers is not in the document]
ferdi05: sorry this is coming out of the blue, I may have missed the notification here for some reasons. anyway happy to discuss if we can make an article. Can we have a quick chat about it today @ManyTheFish @loiclec ? otherwise when I'll be back from vacations in 2 weeks?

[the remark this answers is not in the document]
ManyTheFish: Hello @ferdi05, speaking with @loiclec, there are some improvements that can be done on this part, so, waiting until the end of your vacation seems to be better. 😄

milli/src/update/word_prefix_pair_proximity_docids.rs
```diff
@@ -169,78 +373,549 @@ impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
...
+
+    fn reclaim_byte_vector(&mut self, mut data: Vec<u8>) {
+        data.clear();
+        self.byte_vectors.push(data);
+    }
+    fn reclaim_mergeable_data_vector(&mut self, mut data: Vec<Cow<'static, [u8]>>) {
+        data.clear();
+        self.mergeable_data_vectors.push(data);
+    }
+}
```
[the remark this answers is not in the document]
loiclec: I have removed it. With mimalloc, on my computer, it made little to no difference :))

milli/src/update/word_prefix_pair_proximity_docids.rs
loiclec: done 👍  let me know if the explanation is good enough

infos/src/main.rs
loiclec: I talked a little bit with Many about it, and I think the answer is that it is not strictly needed. I added it there initially so that we can be sure that the keys are lexicographically ordered tuples (word1, word2, proximity).

But the proximity is always very small (= ~3 I think?), and the tokeniser removes all character bytes from the word tokens, so that they cannot contains bytes smaller than 3. So, almost by accident, the lexicographical order is respected anyway.

I would still rather make it explicit though. If we ever accept larger proximities, or character bytes of `0x1, 0x2, 0x3`, then it will cause very confusing bugs.

milli/src/update/word_prefix_pair_proximity_docids.rs
```diff
@@ -169,78 +373,549 @@ impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
...
+/// Its main arguments are:
+/// 1. a sorted iterator over ((word1, word2, proximity), docids) elements
+/// 2. a prefix trie
+/// 3. a closure to describe how to handle the new computed (word1, prefix, proximity) elements
+///
+/// For more information about what this function does, read the module documentation.
+fn execute_on_word_pairs_and_prefixes<Iter>(
+    iter: &mut Iter,
+    mut next_word_pair_proximity: impl for<'a> FnMut(
+        &'a mut Iter,
```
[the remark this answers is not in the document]
loiclec: done :)

Kerollmops (approved):

milli/src/update/word_prefix_pair_proximity_docids.rs
```diff
@@ -169,89 +368,348 @@ impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
...
+The batch is flushed as often as possible, when we are sure that every (word1, prefix, proximity) key derived from its content
+can be inserted into the database in sorted order. When it is flushed, it calls a user-provided closure with the following arguments:
+- key   : (word1, prefix, proximity) as bytes
+- value : merged roaring bitmaps from all values associated with (prefix, proximity) in the batch, serialised to bytes
+*/
+#[derive(Default)]
+struct PrefixAndProximityBatch {
+    word1: Vec<u8>,
+    batch: Vec<(Vec<u8>, Vec<Cow<'static, [u8]>>)>,
+}
```
Kerollmops, line 473: I am thinking about a potential issue with this data structure: could you prove to me that it doesn't consume to much memory? I mean, if you tell me that the tree we not grow much as it is flush every time the word1 changes it is OK.

milli/src/update/word_prefix_pair_proximity_docids.rs
```diff
@@ -169,89 +368,348 @@ impl<'t, 'u, 'i> WordPrefixPairProximityDocids<'t, 'u, 'i> {
...
+╚══════╝╚══════╝╚══════╝ └─────────┘  ╚═════════╝ └─────────┘ ╚══════════╝
+                         ╔═══╗ ╔═══╗                 ╔═══╗
+                         ║ i ║ ║ l ║                 ║ l ║
+                         ╚═══╝ ╚═══╝                 ╚═══╝
+*/
+#[derive(Default, Debug)]
+struct PrefixTrieNode {
+    children: Vec<(PrefixTrieNode, u8)>,
+    is_end_node: bool,
+}
```
Kerollmops, line 605: How much memory will this tree consume in normal use cases? I mean, will it be flush fast enough to not trigger a memory issue.

Kerollmops (changes requested):

milli/src/update/word_prefix_pair_proximity_docids.rs
loiclec: It is flushed every time the word1 changes and every time the first letter of word2 changes, so it never gets big.

milli/src/update/word_prefix_pair_proximity_docids.rs
loiclec: It takes a completely negligible amount of memory and never gets flushed. Reasons:
1. it contains only the prefixes corresponding to more than 50 different words, which are rare
2. the maximum length of the prefixes is very short (typically 2 bytes only)

Kerollmops (approved):

Kerollmops (approved):
After those explanations, I would say that it is good to be merged as it is much faster and doesn't consume too much memory.
bors merge

Kerollmops (approved):
bors merge

PR #588, opened 2022-07-18
Fix name of "release_date" facet in movies benchmarks

loiclec:
## What does this PR do?
The `movies.json` file in the benchmark datasets contains a filterable field called "release_date", but the indexing benchmarks wrongly called the field "released_date" instead. This PR fixes that.


ManyTheFish (approved):
Thank you @loiclec for pointing out this mistake!

bors merge

PR #590, opened 2022-07-20
Optimise facets indexation

### LINKED ISSUE #589: Optimise indexing of facet databases
When creating the `facet_id_string_docids` and `facet_id_f64_docids` databases, we need to perform approximately `n * log_b(n)` roaring bitmap unions, where `n` is the number of different facet values for a certain field in all the documents and `b` is the "level group size". This is because we iterate over the level 0 of these databases (containing all the unique field values) for every additional level we want to create.

For example, in the benchmark `Indexing geo_point`, we have a lot of different documents with unique longitudes and latitudes, which are faceted fields. If there are 1,000,000 of them, then we need to create 8 levels, and thus perform 8,000,000 bitmap unions. This is expensive. 

It is possible to use a recursive algorithm instead which computes level N from the contents of level N-1. Doing so would reduce the number of bitmap unions from `n * log_b(n)` to `n + 1/(b-1) * n` (`n` times for level 1, `n/b` times for level 2, `n/b^2` for level 3, etc.).

Doing so is important because in datasets with a lot of different facet values, a very significant portion of the time is spent computing these bitmaps.

loiclec:
# Pull Request

## What does this PR do?
Fixes #589 

## Notes
I added documentation for the whole module which attempts to explain the shape of the databases and their purpose. However, I realise there is already some documentation about this, so I am not sure if we want to keep it.

## Benchmarks
(Coming soon)


loiclec:
Note: I'm putting this PR back into draft status as I don't think the improvement is good enough.

The problem is that after adding just one document, we need to iterate over the whole database and perform a ton of bitmap operations, which is very slow. Ideally, the algorithm should be incremental and modify the database in-place instead of clearing it and rebuilding it from scratch. I am starting to think about what a proper solution will be and will update this PR later.

loiclec:
After looking a bit more into it, I realised there is a problem with the current structure of `index.facet_id_string_docids`. The elements of all levels > 0 have a pair of indices as key. These indices point to elements of level 0. For example:
```
┌───────┐   ┌───────────────────────────────────────────────────────────────────────────────┐
│  all  │   │                       [a, b, c, d, e, f, g, u, y, z]                          │
└───────┘   └───────────────────────────────────────────────────────────────────────────────┘

            ┌───────────────────────────────┬───────────────────────────────┬───────────────┐
┌───────┐   │             0 – 3             │             4 – 7             │     8 – 9     │
│Level 2│   │                               │                               │               │
└───────┘   │        [a, b, d, f, z]        │        [c, d, e, f, g]        │    [u, y]     │
            ├───────────────┬───────────────┼───────────────┬───────────────┼───────────────┤
┌───────┐   │     0 – 1     │     2 – 3     │     4 – 5     │     6 – 7     │     8 – 9     │
│Level 1│   │  "ab" – "ac"  │ "ba" – "bac"  │ "gaf" – "gal" │"form" – "wow" │ "woz" – "zz"  │
└───────┘   │ [a, b, d, z]  │   [a, b, f]   │   [c, d, g]   │    [e, f]     │    [u, y]     │
            ├───────┬───────┼───────┬───────┼───────┬───────┼───────┬───────┼───────┬───────┤
┌───────┐   │  "ab" │  "ac" │  "ba" │ "bac" │ "gaf" │ "gal" │ "form"│ "wow" │ "woz" │  "zz" │
│Level 0│   │  "AB" │ " Ac" │ "ba " │ "Bac" │ " GAF"│ "gal" │ "Form"│ " wow"│ "woz" │  "ZZ" │
└───────┘   │ [a, b]│ [d, z]│ [b, f]│ [a, f]│ [c, d]│  [g]  │  [e]  │ [e, f]│  [y]  │  [u]  │
            └───────┴───────┴───────┴───────┴───────┴───────┴───────┴───────┴───────┴───────┘
```

As a result, adding an element at level 0 automatically invalidates all its parents and its parents' successors. That means the minimum complexity of adding an element is `Ω(size_level_0)`, which is too high. 

I think this could be solved by using strings directly instead of indices as keys, like this:
```
┌───────┐   ┌───────────────────────────────────────────────────────────────────────────────┐
│  all  │   │                         a, b, c, d, e, f, g, u, y , z                         │
└───────┘   └───────────────────────────────────────────────────────────────────────────────┘
                                                                                             
            ┌───────────────────────────────┬───────────────────────────────┬───────────────┐
┌───────┐   │         "ab" – "bac"          │         "gaf" – "wow"         │ "woz" – "zz"  │
│Level 2│   │                               │                               │               │
└───────┘   │         a, b, d, f, z         │         c, d, e, f, g         │     u, y      │
            ├───────────────┬───────────────┼───────────────┬───────────────┼───────────────┤
┌───────┐   │  "ab" – "ac"  │ "ba" – "bac"  │ "gaf" – "gal" │"form" – "wow" │ "woz" – "zz"  │
│Level 1│   │               │               │               │               │               │
└───────┘   │  a, b, d, z   │    a, b, f    │    c, d, g    │     e, f      │     u, y      │
            ├───────┬───────┼───────┬───────┼───────┬───────┼───────┬───────┼───────┬───────┤
┌───────┐   │  "ab" │  "ac" │  "ba" │ "bac" │ "gaf" │ "gal" │ "form"│ "wow" │ "woz" │  "zz" │
│Level 0│   │  "AB" │ " Ac" │ "ba " │ "Bac" │ " GAF"│ "gal" │ "Form"│ " wow"│ "woz" │  "ZZ" │
└───────┘   │  a, b │  d, z │  b, f │  a, f │  c, d │   g   │   e   │  e, f │   y   │   u   │
            └───────┴───────┴───────┴───────┴───────┴───────┴───────┴───────┴───────┴───────┘
```

But as explained in `search/facet/facet_string.rs`, this would result in very long keys, and we would need to reduce the maximum allowed size of facet strings by half, to approximately 250 bytes (the max key size of LMDB is 511 bytes).

loiclec:
After thinking about it more, I think this PR should be reviewed as is for now, because:
1. It still brings a fairly large improvement in indexing speed
2. Adding code to incrementally update the database will be complicated and take a long time
3. Even after we have the ability to incrementally update the DB, we will still need to use this bit of code to build the DB from scratch

milli/src/update/facets.rs
```diff
@@ -151,178 +260,332 @@ fn compute_facet_number_levels<'t>(
...
+    write_entry: &dyn Fn(&mut Writer<File>, NonZeroU8, Bound, Bound, RoaringBitmap) -> Result<()>,
+) -> Result<Vec<Reader<File>>>
+where
+    KeyCodec: for<'a> BytesEncode<'a>
+        + for<'a> BytesDecode<'a, DItem = <KeyCodec as BytesEncode<'a>>::EItem>,
+    for<'a> <KeyCodec as BytesEncode<'a>>::EItem: Sized,
+    ValueCodec: for<'a> BytesEncode<'a>
+        + for<'a> BytesDecode<'a, DItem = <ValueCodec as BytesEncode<'a>>::EItem>,
+    for<'a> <ValueCodec as BytesEncode<'a>>::EItem: Sized,
+    Bound: Copy,
```
Kerollmops, line 445: This is one of the reasons why I would like to rethink the `ByteEncode/ByteDecode` trait to be simpler, if possible, and be directly implemented on the serialized types instead of serializer types ([kind of like sanakirja](https://docs.rs/sanakirja/latest/sanakirja/trait.Storable.html)).

milli/src/update/facets.rs
```diff
@@ -151,178 +260,332 @@ fn compute_facet_number_levels<'t>(
...
+fn compute_facet_strings_levels<'t>(
     rtxn: &'t heed::RoTxn,
     db: heed::Database<FacetStringLevelZeroCodec, FacetStringLevelZeroValueCodec>,
     compression_type: CompressionType,
     compression_level: Option<u32>,
     level_group_size: NonZeroUsize,
     min_level_size: NonZeroUsize,
     field_id: FieldId,
-) -> Result<Reader<File>> {
+) -> Result<(Vec<Reader<File>>, RoaringBitmap)> {
```
Kerollmops, line 327: Could you describe what's the output of this function, now?

Kerollmops (changes requested):

milli/src/update/facets.rs
loiclec: I would also like to rethink the `BytesEncode/BytesDecode` traits :-)

Kerollmops (approved):

Kerollmops (approved):
Thank you very much!
bors merge

PR #592, opened 2022-07-20
Fix(smart-crop): Ensure that matches are all highlighted

ManyTheFish:
In the case of only 1 query word was matching in an attribute,
and this same word was matching several times in the crop window of the same attribute,
then only the first match was highlighted.
Now we ensure that the computed "best matches interval" contains all the words that match in the same crop window.

related to https://github.com/meilisearch/meilisearch/issues/2627

### DIFF
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "benchmarks"
-version = "0.31.1"
+version = "0.31.2"
 edition = "2018"
 publish = false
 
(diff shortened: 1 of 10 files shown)


curquiza (changes requested):
Don't forget the label 😄 

Also, can you change the version name in this PR at the same time?

curquiza (approved):
Good for me, but I'm not the one who is able to say the code is ok 😄

loiclec (approved):
Thanks, it looks good!

PR #594, opened 2022-07-21
Fix(Search): Fix phrase search candidates computation

ManyTheFish:
This bug is an old bug but was hidden by the proximity criterion,
Phrase searches were always returning an empty candidates list when the proximity criterion is deactivated.

Before the fix, we were trying to find any words[n] near words[n]
instead of finding  any words[n] near words[n+1], for example:

for a phrase search '"Hello world"' we were searching for "hello" near "hello" first, instead of "hello" near "world".

### DIFF
--- a/milli/src/search/criteria/mod.rs
+++ b/milli/src/search/criteria/mod.rs
@@ -293,13 +293,13 @@ impl<'t> CriteriaBuilder<'t> {
     }
 }
 
-pub fn resolve_query_tree<'t>(
-    ctx: &'t dyn Context,
+pub fn resolve_query_tree(
+    ctx: &dyn Context,
     query_tree: &Operation,
     wdcache: &mut WordDerivationsCache,
 ) -> Result<RoaringBitmap> {
-    fn resolve_operation<'t>(
-        ctx: &'t dyn Context,
+    fn resolve_operation(
+        ctx: &dyn Context,
         query_tree: &Operation,
         wdcache: &mut WordDerivationsCache,
     ) -> Result<RoaringBitmap> {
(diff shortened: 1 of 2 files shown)


milli/src/search/criteria/proximity.rs
```diff
@@ -192,22 +192,42 @@ fn resolve_candidates<'t>(
...
+
+                        for bitmap in bitmaps {
+                            if first_iter {
+                                candidates = bitmap;
+                                first_iter = false;
+                            } else {
+                                candidates &= bitmap;
+                            }
+                            // There will be no match, return early
+                            if candidates.is_empty() {
```
ManyTheFish, line 224: This is a copy-paste of the other modified file, we didn't have the same behavior in proximity as in the main function, this difference was hiding the bug located in the main function.

loiclec: Should we maybe factor it out in a separate function if we want the two copies to stay the same? It's likely that they will diverge over time if we are not careful. 

As an alternative, add a comment linking the two copies together, something like:
```rust
// The code below is exactly the same as #hGVU7jMhU4Mwr4XoU77B
let mut candidates = RoaringBitmap::new();


// and for the second copy as well:
// The code below is exactly the same as #hGVU7jMhU4Mwr4XoU77B
let mut candidates = RoaringBitmap::new();

```

irevoire: > Should we maybe factor it out in a separate function if we want the two copies to stay the same?

That would be cool if it’s possible

milli/src/search/criteria/mod.rs
```diff
@@ -378,6 +342,44 @@ pub fn resolve_query_tree<'t>(
     resolve_operation(ctx, query_tree, wdcache)
 }
 
+pub fn resolve_phrase<'t>(ctx: &'t dyn Context, phrase: &[String]) -> Result<RoaringBitmap> {
```
Kerollmops, line 345:
```suggestion
pub fn resolve_phrase(ctx: &dyn Context, phrase: &[String]) -> Result<RoaringBitmap> {
```

Kerollmops (changes requested):

irevoire (dismissed):
looks good

Kerollmops (approved):
bors merge

PR #596, opened 2022-07-21
Filter operators: NOT + IN[..]

loiclec:
# Pull Request

## What does this PR do?
Implements the changes described in https://github.com/meilisearch/meilisearch/issues/2580
It is based on top of #556


irevoire (changes requested):
noice

filter-parser/src/lib.rs
```diff
@@ -170,36 +164,119 @@ fn ws<'a, O>(inner: impl FnMut(Span<'a>) -> IResult<O>) -> impl FnMut(Span<'a>)
     delimited(multispace0, inner, multispace0)
 }
 
-/// or             = and ("OR" WS+ and)*
+/// value_list = (value ("," value)* ","?)?
+fn parse_value_list<'a>(input: Span<'a>) -> IResult<Vec<Token<'a>>> {
+    // TODO: here, I should return a failure with a clear explanation whenever possible
+    // for example:
```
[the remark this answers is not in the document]
loiclec: I added better errors now :)

milli/src/search/facet/filter.rs
```diff
@@ -732,12 +740,10 @@ mod tests {
...
-        let error = Filter::from_str(&filter_string).unwrap_err();
-        assert!(
-            error.to_string().starts_with("Too many filter conditions"),
-            "{}",
-            error.to_string()
-        );
+        // Note: the filter used to be rejected for being too deep, but that is
+        // no longer the case
+        let filter = Filter::from_str(&filter_string).unwrap();
+        assert!(filter.is_some());
```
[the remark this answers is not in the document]
loiclec: I fixed that case as well, and I added a filter depth limitation to stop stack overflows. However, I don't think the end user should worry about it. There are no realistic filters that would reach the limit.

filter-parser/src/lib.rs
```diff
@@ -292,371 +401,329 @@ pub mod tests {
...
+            @"{x} = {1}"
+        );
+        insta::assert_display_snapshot!(
+            p!("NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT NOT x = 1"),
+            @"NOT ({x} = {1})"
+        );
+
+        // Confusing keywords
+        insta::assert_display_snapshot!(p!(r#"NOT "OR" EXISTS AND "EXISTS" NOT EXISTS"#), @"AND[NOT ({OR} EXISTS), NOT ({EXISTS} EXISTS), ]");
```
irevoire, line 492: what a bunch of perfectly fine little requests

filter-parser/src/lib.rs
```diff
@@ -292,371 +401,329 @@ pub mod tests {
...
-                "Filter `{}` wasn't supposed to be parsed but it did with the following result: `{:?}`",
-                input,
-                result.unwrap()
-            );
-            let filter = result.unwrap_err().to_string();
-            assert!(filter.starts_with(expected), "Filter `{:?}` was supposed to return the following error:\n{}\n, but instead returned\n{}\n.", input, expected, filter);
+        macro_rules! p {
+            ($s:literal) => {
+                Fc::parse($s).unwrap_err().to_string()
+            };
```
irevoire, line 502: Why this isn't a function?

filter-parser/src/lib.rs
```diff
@@ -292,371 +401,329 @@ pub mod tests {
...
+                    write!(f, "{el}, ")?;
+                }
+                write!(f, "]")
+            }
+            FilterCondition::GeoLowerThan { point, radius } => {
+                write!(f, "_geoRadius({}, {}, {})", point[0], point[1], radius)
+            }
+        }
+    }
+}
```
irevoire, line 710: Is there a reason why you prefer to reimplement Display instead of using the Debug default implementation?

We're losing the `Span` info, which could be cool to check, right?

milli/src/search/facet/filter.rs
```diff
@@ -89,52 +89,44 @@ impl<'a> Filter<'a> {
...
-                        ands = match ands.take() {
-                            Some(ands) => {
-                                Some(FilterCondition::And(Box::new(ands), Box::new(rule)))
-                            }
-                            None => Some(rule),
-                        };
+                    if ors.len() > 1 {
+                        ands.push(FilterCondition::Or(ors));
+                    } else if ors.len() == 1 {
+                        ands.push(ors[0].clone());
```
irevoire, line 107:
```suggestion
                        ands.push(ors[0].pop().unwrap());
```

Shouldn't this work?

milli/src/search/facet/filter.rs
```diff
@@ -89,52 +89,44 @@ impl<'a> Filter<'a> {
...
-                        };
+                        ands.push(filter.condition);
                     }
                 }
             }
         }
+        let and = if ands.is_empty() {
+            return Ok(None);
+        } else if ands.len() == 1 {
+            ands[0].clone()
```
irevoire, line 120:
```suggestion
            ands.pop().unwrap()
```
same

irevoire (changes requested):

filter-parser/src/lib.rs
loiclec: no reason at all haha, it started more complicated and then when I simplified it I kept it as a macro, I'll change it :)

filter-parser/src/lib.rs
loiclec: We're never doing anything with the `Span` information though, right? I wanted to have something that I could check the validity of easily, and the spans were distracting + took a lot of space. 

For error messages where `Span` may be used, we have separate snapshot tests.

filter-parser/src/lib.rs
irevoire: there was another one above that I forgot to comment!

filter-parser/src/lib.rs
irevoire: Yeah, it's only used for the errors.

> I wanted to have something that I could check the validity of easily, and the spans were distracting + took a lot of space.

Fair enough

irevoire (approved):
FINALLY

bors merge

Issue #597, opened 2022-07-21
Write snapshot tests


loiclec:
When refactoring code in milli, it can be hard to make sure that the changes we make are correct.
I experienced this when I worked on https://github.com/meilisearch/milli/pull/587 

1. I refactored a file to improve its performance
2. I ran all of milli's tests, which succeeded
3. I ran the benchmarks, which didn't raise any red flags
4. I even created a new test DB, added some documents into it, and queried it to see if the search results still seemed correct

Everything seemed okay. But then I added a print statement in both the old and new version of the file, to check how many elements were added to the database. By doing so, I realised that I was not adding the same number of elements as before, and therefore my refactor was not faithful to the original file.

Afterwards, this sort of manual differential testing is what gave me the most confidence in my refactoring. Eventually, after fixing a few bugs and triple-checking my code for correctness, I realised the old and new versions were still not the same. So I went further and printed the value of all the elements that were added to the database into two files (old and new). Then, I manually compared the content of those files by putting them side-by-side in Sublime Text. This is how I discovered that I had accidentally fixed a bug (now described in the PR) 😄 

What I'd like to do is to establish a snapshot testing suite to automate the process I went through. The goal of the suite would be to:
1. In the snapshot test, run as many possible operations as possible:
    * using small workloads that are stress-testing edge and corner cases
    * but also using big, real-life workloads, like the ones used in benchmarks
    * these operations are not only related to indexing, but also to search queries, filters, smart crops, etc.
2. And then capture the "outputs" of those operations:
    * the output is the user-visible result of an operation
    * but we can additionally capture lower-level details, like the content of the databases
    * each output is associated with a unique ID that is a combination of the test name and of what the output represents. For example: `output_id = ("movies.json indexing with default values", "content of the word prefix database")`
    * importantly, each output is **untyped**. For example, it can be a serialised JSON response, or the raw key and value bytes in the database.
3. Finally, compare each output to its previous state before the latest changes.
    * Assert that the contents are equal.
    * If they are not equal, explain exactly what changed. For example: `("movies.json indexing with default values", "content of the word prefix database") changed. Previous: <path to old output file>, New: <path to new output file>`.
    * And then give us the possibility to validate these changes. This is necessary when we willingly change the behaviour of the code. When we validate a change, the previous output file is replaced by the new one.

[...]

PR #601, opened 2022-08-03
Introduce snapshot tests

loiclec:
# Pull Request
## What does this PR do?
Introduce snapshot tests into milli, by using the `insta` crate. This implements the idea described by #597 

See: [insta.rs](https://insta.rs)

## Design
There is now a new file, `snapshot_tests.rs`, which is compiled only under `#[cfg(test)]`. It exposes the `db_snap!` macro, which is used to snapshot the content of a database.

When running `cargo test`, `insta` will check that the value of the current snapshot is the same as the previous one (on the file system). If they are the same, the test passes. If they are different, the test fails and you are asked to review the new snapshot to approve or reject it.

We don't want to save very large snapshots to the file system, because it will pollute the git repository and increase its size too much. Instead, we only save their `md5` hashes under the name `<snapshot_name>.hash.snap`. There is a new environment variable called `MILLI_TEST_FULL_SNAPS` which can be set to `true` in order to *also* save the full content of the snapshot under the name `<snapshot_name>.full.snap`. However, snapshots with the extension `.full.snap` are never saved to the git repository.

## Example
```rust
// In e.g. facets.rs
#[test]
fn my_test() {
    // create an index
    let index = TempIndex::new():
    index.add_documents(...);
    index.update_settings(|settings| ...);
    
    // then snapshot the content of one of its databases
    // the snapshot will be saved at the current folder under facets.rs/my_test/facet_id_string_docids.snap
    db_snap!(index, facet_id_string_docids);

    index.add_documents(...);   

    // we can also name the snapshot to ensure there is no conflict
    // this snapshot will be saved at facets.rs/my_test/updated/facet_id_string_docids.snap
    db_snap!(index, facet_id_string, docids, "updated");
    
    // and we can also use "inline" snapshots, which insert their content in the given string literal
    db_snap!(index, field_distributions, @"");
    // once the snapshot is approved, it will automatically get transformed to, e.g.:
    // db_snap!(index, field_distributions, @"
    // my_facet        21
    // other_field     3
    // ");
    
    // now let's add **many** documents
    index.add_documents(...);
    
    // because the snapshot is too big, its hash is saved instead
    // if the MILLI_TEST_FULL_SNAPS env variable is set to true, then the full snapshot will also be saved
    // at facets.rs/my_test/large/facet_id_string_docids.full.snap
    db_snap!(index, facet_id_string_docids, "large", @"5348bbc46b5384455b6a900666d2a502");
}
```


milli/src/search/query_tree.rs
```diff
@@ -857,532 +852,263 @@ mod test {
...
         let (query_tree, _) =
             TestContext::default().build(false, true, None, tokens).unwrap().unwrap();
 
-        assert_eq!(expected, query_tree);
+        insta::assert_debug_snapshot!(query_tree, @r###"
+        OR
+          AND
+            Exact { word: "hey" }
+            PrefixTolerant { word: "friends", max typo: 1 }
+          PrefixTolerant { word: "heyfriends", max typo: 1 }
```
Kerollmops, line 863: We must make sure that this debug-printing is reliable and that the whole set of fields is displayed.

Kerollmops (changes requested):
It is pretty amazing the amount of lines of code that you removed by defining new helping methods on the `TestIndex` struct 🎉

Thank you very much for introducing insta.rs in milli, it looks amazing, I am impatient to learn more about it!

irevoire (dismissed):
That’s super noice, can’t wait to use it!

loiclec:
Does anybody know why the tests are failing on Windows? 🤔

milli/src/search/query_tree.rs
loiclec: Yep, I checked it :)

Kerollmops:
> Does anybody know why the tests are failing on Windows? 🤔

I have no clue. It looks like it is failing to infer types but ONLY on windows? Is it due to the new version of Rust?

Kerollmops (approved):
Looks good to me, thank you very much!
bors merge

PR #602, opened 2022-08-10
Use mimalloc as the default allocator

loiclec:
## What does this PR do?
Use mimalloc as the global allocator for milli's benchmarks on macOS.

## Why?
On Linux, we use jemalloc, which is a very fast allocator. But on macOS, we currently use the system allocator, which is very slow. In practice, this difference in allocator speed means that it is difficult to gain insight into milli's performance by running benchmarks locally on the Mac.

By using mimalloc, which is another excellent allocator, we reduce the speed difference between the two platforms.


irevoire:
From what I see mimalloc is almost always faster than jemalloc, why not use it for linux as well?

benchmarks/Cargo.toml
```diff
@@ -13,6 +13,10 @@ csv = "1.1.6"
 [target.'cfg(target_os = "linux")'.dependencies]
 jemallocator = "0.3.2"
 
+[target.'cfg(target_os = "macos")'.dependencies]
+mimalloc = { version = "0.1.29", default-features = false }
+
+
```
Kerollmops, line 19:
```suggestion
```

Kerollmops (changes requested):
Indeed, why not using either jemalloc or mimalloc on all platforms. And if we use mimalloc it could probably also work on Windows too (as it is made by Microsoft)?

benchmarks/Cargo.toml
loiclec: I am also happy with using mimalloc everywhere. I wanted to make a minimal change in the first place, which is why I kept jemalloc on Linux. Using jemalloc everywhere wouldn't be possible I think, since according to https://lib.rs/crates/jemallocator , some jemalloc tests don't pass on macOS.

loiclec: What I can do is:
1. Amend the PR to unconditionally use mimalloc as the global allocator everywhere
2. Run the benchmarks

Then:
1. If the benchmarks did not regress, merge the PR, then use mimalloc on the meilisearch repo as well
2. If the benchmarks regressed, only use mimalloc on macOS

Does that sound good?

Kerollmops: Yup, sound very good to me :)

loiclec:
It looks like mimalloc is indeed faster than jemalloc, so I have enabled it on all platforms now.

```
group                                                                     indexing_main_950d8e4c                 indexing_mimalloc-benchmarks_fb2b6c0c
-----                                                                     ----------------------                 -------------------------------------
indexing/-geo-delete-facetedNumber-facetedGeo-searchable-                 1.00     35.6±2.96ms        ? ?/sec    1.09     38.7±4.25ms        ? ?/sec
indexing/-movies-delete-facetedString-facetedNumber-searchable-           1.00      8.3±2.72ms        ? ?/sec    1.33     11.1±5.47ms        ? ?/sec
indexing/-movies-delete-facetedString-facetedNumber-searchable-nested-    1.05     12.7±2.56ms        ? ?/sec    1.00     12.1±1.57ms        ? ?/sec
indexing/-songs-delete-facetedString-facetedNumber-searchable-            1.01     42.7±6.45ms        ? ?/sec    1.00     42.3±4.80ms        ? ?/sec
indexing/-wiki-delete-searchable-                                         1.05   283.0±18.79ms        ? ?/sec    1.00   268.9±22.20ms        ? ?/sec
indexing/Indexing geo_point                                               1.07      66.0±0.40s        ? ?/sec    1.00      61.9±0.65s        ? ?/sec
indexing/Indexing movies in three batches                                 1.07      16.7±0.18s        ? ?/sec    1.00      15.6±0.09s        ? ?/sec
indexing/Indexing movies with default settings                            1.07      14.0±0.30s        ? ?/sec    1.00      13.0±0.27s        ? ?/sec
indexing/Indexing nested movies with default settings                     1.05      10.8±0.18s        ? ?/sec    1.00      10.3±0.31s        ? ?/sec
indexing/Indexing nested movies without any facets                        1.06       9.6±0.11s        ? ?/sec    1.00       9.0±0.09s        ? ?/sec
indexing/Indexing songs in three batches with default settings            1.06      67.4±1.16s        ? ?/sec    1.00      63.5±0.61s        ? ?/sec
indexing/Indexing songs with default settings                             1.08      59.9±1.21s        ? ?/sec    1.00      55.7±0.58s        ? ?/sec
indexing/Indexing songs without any facets                                1.07      54.4±1.01s        ? ?/sec    1.00      50.9±0.58s        ? ?/sec
indexing/Indexing songs without faceted numbers                           1.06      58.9±1.21s        ? ?/sec    1.00      55.4±1.23s        ? ?/sec
indexing/Indexing wiki                                                    1.05   1077.7±25.72s        ? ?/sec    1.00   1022.7±11.20s        ? ?/sec
indexing/Indexing wiki in three batches                                   1.04    1193.7±7.22s        ? ?/sec    1.00    1143.3±4.03s        ? ?/sec
indexing/Reindexing geo_point                                             1.04      25.6±0.08s        ? ?/sec    1.00      24.5±0.14s        ? ?/sec
indexing/Reindexing movies with default settings                          1.00    335.8±9.11ms        ? ?/sec    1.00   334.6±29.84ms        ? ?/sec
indexing/Reindexing songs with default settings                           1.08       7.2±0.14s        ? ?/sec    1.00       6.7±0.08s        ? ?/sec
indexing/Reindexing wiki                                                  1.03   1750.9±28.38s        ? ?/sec    1.00   1693.3±29.61s        ? ?/sec
```

Kerollmops (commented):
I am wondering why we are only using mimalloc for the benchmarks and not everything.

helpers/Cargo.toml
```diff
@@ -14,5 +14,4 @@ milli = { path = "../milli" }
 stderrlog = "0.5.1"
 structopt = { version = "0.3.26", default-features = false }
 
-[target.'cfg(target_os = "linux")'.dependencies]
-jemallocator = "0.3.2"
+mimalloc = { version = "0.1.29", default-features = false }
```
Kerollmops, line 17: Could you add the newline here, please?
FYI there is a setting for that in sublime text:
```json
{ "ensure_newline_at_eof_on_save": true }
```

Kerollmops (changes requested):

helpers/Cargo.toml
loiclec: yes, thanks for the setting! it's set now 😄

loiclec:
mimalloc is used for everything now. Well, almost. It could also be used as a dev-dependency for milli's tests I guess. But I don't think it is very important, and I am afraid of having conflicting global allocators somehow.

helpers/Cargo.toml
```diff
@@ -14,5 +14,4 @@ milli = { path = "../milli" }
 stderrlog = "0.5.1"
 structopt = { version = "0.3.26", default-features = false }
```
Kerollmops, line 16:
```suggestion
```

Kerollmops (changes requested):
Could you, please, sort the _Cargo.toml_ dependencies?

loiclec:
Yes, done :-)

Kerollmops (approved):
Thank you very much!
bors merge

PR #606, opened 2022-08-17
Make binaries faster on release profile through better compile options

loiclec:
Using `codegen-units = 1` and `lto = 'thin'` makes the compile time a bit longer, but also produces faster binaries.

I'd like to run milli's benchmark with these options, so that we can see whether it is worth enabling on meilisearch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,8 @@ opt-level = 3
 
 [profile.release]
 debug = true
+codegen-units = 1
+lto = "thin"
 
 # Make sure that the build scripts and proc-macros are compiled with
 # all the optimizations. It speeds up the zip crate that we use in the build.rs.


Cargo.toml
```diff
@@ -8,6 +8,9 @@ opt-level = 3
 
 [profile.release]
 debug = true
+codegen-units = 1
+lto = "thin"
+
```
Kerollmops, line 13:
```suggestion
```

Kerollmops (changes requested):
Do these options matter for a library?

loiclec:
They are not read when milli is compiled as a library. But for all the executables that are part of the workspace (benchmarks, cli, etc.), it should make things a bit faster. If we decide to add them to meilisearch, then I think it's best to have them enabled for the benchmarks as well I think.

Kerollmops (approved):
bors merge

PR #607, opened 2022-08-17
Better threshold

### LINKED ISSUE #570: Choose a better threshold for the soft deletion
Once #557 is merged the engine will wait until 10k documents have been soft deleted before running a real deletion.
Since this number is so smol it should not cause any issue, but the bigger this number is and the bigger the perf improvement is. BUT if this number gets too big we're going to cancel indexing tasks saying there is no space left on the device when in reality there are a lot of documents waiting to be deleted.

Thus I would like to think of a better threshold, currently, my favourite idea would be to calculate an approximative document size (we take the size of the DB divided by the number of documents + soft-deleted documents) and, if the number of soft-deleted documents reaches 10% of the total space we can use then we delete them for real.
Also maybe this could be configurable. Maybe instead of 10% we should instead say something like « 100mb » (but in this case it should really be configurable).

irevoire:
# Pull Request

## What does this PR do?
Fixes #570 

This PR tries to improve the threshold used to trigger the real deletion of documents.
The deletion is now triggered in two cases;
- 10% of the total available space is used by soft deleted documents
- 90% of the total available space is used.

In this context, « total available space » means the `map_size` of lmdb.
And the size used by the soft deleted documents is actually an estimation. We can't determine precisely the size used by one document thus what we do is; take the total space used, divide it by the number of documents + soft deleted documents to estimate the size of one average document. Then multiply the size of one avg document by the number of soft deleted document.

--------

image

Here we can see we have a ~10GB drift in the end between the space used by the soft deleted and the real space used by the documents.
Personally I don’t think that's a big issue because once the red line reach 90GB everything will be freed but now you know.

If you have an idea on how to improve this estimation I would love to hear it.
It look like the difference is linear so maybe we could simply multiply the current estimation by two?

### DIFF
--- a/benchmarks/benches/indexing.rs
+++ b/benchmarks/benches/indexing.rs
@@ -5,7 +5,7 @@ use std::fs::{create_dir_all, remove_dir_all};
 use std::path::Path;
 
 use criterion::{criterion_group, criterion_main, Criterion};
-use heed::{EnvOpenOptions, RwTxn};
+use milli::heed::{EnvOpenOptions, RwTxn};
 use milli::update::{
     DeleteDocuments, IndexDocuments, IndexDocumentsConfig, IndexerConfig, Settings,
 };
--- a/benchmarks/benches/utils.rs
+++ b/benchmarks/benches/utils.rs
@@ -6,8 +6,8 @@ use std::num::ParseFloatError;
 use std::path::Path;
 
 use criterion::BenchmarkId;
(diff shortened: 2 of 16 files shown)


Kerollmops (approved):

Kerollmops (approved):
Thank you very much! Looks good and well commented!
bors merge

Kerollmops:
bors merge

PR #609, opened 2022-08-17
Retry downloading the benchmarks datasets

irevoire:
Downloading the benchmarks datasets is failing [more and more](https://github.com/meilisearch/milli/pull/607#pullrequestreview-1076023074) often; thus, instead of fixing the issue, I thought we could retry multiple times.

### DIFF
--- a/benchmarks/build.rs
+++ b/benchmarks/build.rs
@@ -80,13 +80,22 @@ fn main() -> anyhow::Result<()> {
         }
         let url = format!("{}/{}.{}.gz", BASE_URL, dataset, extension);
         eprintln!("downloading: {}", url);
-        let bytes = download_dataset(url.clone())?;
+        let bytes = retry(|| download_dataset(url.clone()), 10)?;
         eprintln!("{} downloaded successfully", url);
         eprintln!("uncompressing in {}", out_file.display());
         uncompress_in_file(bytes, &out_file)?;
     }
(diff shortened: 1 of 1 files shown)


benchmarks/build.rs
```diff
@@ -89,6 +89,15 @@ fn main() -> anyhow::Result<()> {
     Ok(())
 }
 
+fn retry<Ok, Err>(fun: impl Fn() -> Result<Ok, Err>, times: usize) -> Result<Ok, Err> {
+    for _ in 0..times {
+        if let ok @ Ok(_) = fun() {
+            return ok;
+        }
+    }
```
Kerollmops, line 97: Don't we want to sleep, here?

Kerollmops (changes requested):

benchmarks/build.rs
irevoire: As you want, I don't kwow why it’s failing

Kerollmops: I neither know 😂

irevoire:
Soooo... Should we merge, or should I sleep?
I would say we can merge it like that, and if it doesn't solve anything, then we add a sleep in it 👀

Kerollmops (approved):
Indeed, let's merge and see if the problem persists!
bors merge

PR #610, opened 2022-08-19
Share heed between all sub-crates

irevoire:
# Pull Request

## What does this PR do?
Use the reexported version of heed in the benchmarks and the fuzzer


Kerollmops (approved):
bors merge

PR #615, opened 2022-08-23
Remove the artifacts of the past

irevoire:


Kerollmops (approved):
Wow, that's indeed an ancient relic!
bors merge

PR #616, opened 2022-08-30
Introduce an indexation abortion function when indexing documents

Kerollmops:


loiclec (commented):
Very nice :-)

I worry a bit that compile times will get excessive if we add more closure generic parameters to the indexing functions. But we'll see if that's the case later, and it shouldn't fundamentally change anything to make the closure parameter a dynamic object.

I think we could use the `should_abort` parameter in a lot more places as well. If I recall correctly, the two slowest parts of indexing are the extraction of word pairs (`milli/src/update/index_documents/extract/extract_word_pair_proximity_docids.rs`) and the update of the prefix database (in `milli/src/update/index_documents/mod.rs`). It would be good to abort there as well when possible.

loiclec:
Do you know if it's possible to abort during the extraction of word pairs (when the code in`milli/src/update/index_documents/extract/extract_word_pair_proximity_docids.rs` is executed)? I know we split the dataset into chunks, so maybe we could abort after extracting the data from each chunk there?

loiclec (approved):

irevoire (approved):
it looks good to me 😁

Kerollmops:
bors merge

PR #617, opened 2022-08-31
Accept integers as document ids again

Kerollmops:
This PR is related to https://github.com/meilisearch/meilisearch/issues/2723 and will fix when this PR will be merged, a new release deployed and used in Meilisearch itself.

This PR makes the indexer to try to parse the values of the fields identified as numbers i.e. `id:number` as integer first then as float if it fails.


irevoire (approved):
Thank you for the fix & test!
bors merge

Issue #605, opened 2022-08-15
Improve speed of incremental indexing


loiclec:
Currently there are a few bottlenecks preventing milli to quickly add a document to an existing index. The goal of this issue is to track these bottlenecks.

- [ ]  Facets indexing is not incremental at all, as explained in https://github.com/meilisearch/milli/pull/590 .
- [ ] `word_prefix_docids` and `word_prefix_position_docids` iterate over their whole database to check for elements that need to be deleted, even in the case where we know in advance that nothing has to be deleted.
- [ ] The FST, in general, is not a data structure that can be easily updated. Therefore, `WordsPrefixesFst::execute` as well as the handler for `TypedChunk::WordDocIds` are not incremental either.
- [ ] The RTree used for geosearch can be updated incrementally, but it does need to be deserialised in its entirety first. It is hard to see how else it could work, but I'll mention it here anyway.

As an example, on the `geo_point` dataset, adding a single document takes 3.07s.
- `Facets::execute` takes about 1.55s (**after** https://github.com/meilisearch/milli/pull/590 is implemented, much much longer otherwise)
- `WordPrefixPositionDocids::execute()` and `WordPrefixDocids::execute()` together take 0.2s
- `WordsPrefixesFst::execute()` takes 0.4s
- Handling `TypedChunk::WordDocids` takes 0.73s
- Handling `TypedChunk::GeoPoints` takes 0.17s

All together, these non-incremental parts take about 99.3% of the incremental indexing time.

PR #619, opened 2022-09-01
Refactor the Facets databases to enable incremental indexing

loiclec:
# DRAFT Pull Request

## What does this PR do?
Party fixes https://github.com/meilisearch/milli/issues/605 by making the indexing of the facet databases (i.e. `facet_id_f64_docids` and `facet_id_string_docids`) incremental.

## Remaining tasks
- [ ] Add a lot of documentation about the new structure of the databases as well as the algorithms operating on them
- [ ] Since the `string` and `number` facet databases now share the same structure, there may be more opportunity to deduplicate code. Investigate this.
- [ ] The `infos/` and `http-ui/`crates compile, but they are probably broken. Fix them :)
- [ ] The indexing of the databases should not always be incremental. Find the conditions under which it is faster to recreate them from scratch and dispatch the correct indexing function accordingly.
- [ ] Add more tests:
    - on incremental and non-incremental deletion (be careful to avoid soft-deletion in the tests)
    - on the asc/desc criterion
    - on facet distributions
- [ ] Reorganise the new snapshot tests to use milli's snapshot directory structure
- [ ] Evaluate whether it is worth using fuzzcheck (I used it in the crate I used to develop these changes, and it was useful, but we might not want it in milli)
- [ ] Benchmarking:
   - Compare speed between incremental and from-scratch indexing using the regular benchmarks
   - Add benchmarks for:
       - Adding/Deleting one document to a large dataset
       - Adding/Deleting many documents to a large dataset → use these results to decide when to use incremental indexing 
   - Search: 
      - asc/desc criterion
      - facet distribution
      - filter range search
- [ ] Clean up the git history a little bit (no more `WIP`)
- [ ] Change some of the silly type names I used (e.g. `MyByteSlice`)
- [ ] more, probably


irevoire:
Adding API-breaking for this part;
> Change of behaviour for facet distributions

milli/Cargo.toml
```diff
@@ -54,7 +54,10 @@ big_s = "1.0.2"
 insta = "1.18.1"
 maplit = "1.0.2"
 md5 = "0.7.0"
-rand = "0.8.5"
+rand = {version = "0.8.5", features = ["small_rng"] }
+
+[target.'cfg(fuzzing)'.dev-dependencies]
+fuzzcheck = { path = "../../fuzzcheck-rs/fuzzcheck" }
```
irevoire, line 60: before merging, we shoudl import it from crates.io or github

.gitignore
```diff
@@ -11,6 +13,8 @@
 # Snapshots
 ## ... large
 *.full.snap
-
-#  ... unreviewed
+## ... unreviewed
```
Kerollmops, line 16: Why this comment exist and has been moved?

milli/Cargo.toml
```diff
@@ -54,7 +54,10 @@ big_s = "1.0.2"
 insta = "1.21.0"
 maplit = "1.0.2"
 md5 = "0.7.0"
-rand = "0.8.5"
+rand = {version = "0.8.5", features = ["small_rng"] }
+
+[target.'cfg(fuzzing)'.dev-dependencies]
+fuzzcheck = { git = "https://github.com/loiclec/fuzzcheck-rs", branch = "main" } # TODO: use released version
```
Kerollmops, line 60: Can we use the released version here, please?

milli/Cargo.toml
```diff
@@ -54,7 +54,10 @@ big_s = "1.0.2"
 insta = "1.21.0"
 maplit = "1.0.2"
 md5 = "0.7.0"
-rand = "0.8.5"
+rand = {version = "0.8.5", features = ["small_rng"] }
```
Kerollmops, line 57: Why have we enabled this feature?

Kerollmops (changes requested):
- Why does `http-ui/src/main.rs` and `infos/src/main.rs` exists?

milli/Cargo.toml
loiclec: In tests I want a random number generator that I can give a seed to, and I *think* I had to use `small_rng` for that.

.gitignore
loiclec: the headers were not consistent. It is supposed to:
- Level 1: Snapshots
  - level 2: large
  - level 2: unreviewed
  
But instead, it was:
- Level 1: Snapshots
   - level 2: large
- Level 1: unreviewed

loiclec:
> * Why does `http-ui/src/main.rs` and `infos/src/main.rs` exists?

aftermath of a rebase, they're gone now :)

Kerollmops (dismissed):
Thank you very much! This is the last day and hour of this magnificent PR 👏
It will improve the facet insertion and clean up a lot of code from milli ❤️

bors merge

Kerollmops (approved):
bors merge

curquiza:
bors merge

PR #630, opened 2022-09-07
Prevent windows and milli CI to run for PR

curquiza:
windows and macos GitHub runner (provided by GitHub) are limited, and often we have to wait for the GitHub CI for the PR to have run before CI for bors merging the PR can start.


curquiza:
bors try

curquiza:
Closing this because I've not found any smart way to do it, since the `matrix.os` is not available for a `if` condition on the job level. I would need to duplicate the workflow content, which is not something I would like to do now.

I'm open to any better solution of course 😇

PR #636, opened 2022-09-13
Remove unused `infos`, `http-ui`, and `milli/fuzz`, crates

loiclec:
We haven't used the `infos/`, `http-ui/` and `milli/fuzz/` crates in a long time. They are not properly maintained and probably do not work correctly anymore.

This PR removes these crates entirely from the workspace to reduce the amount of code we need to maintain.

PR #641, opened 2022-09-21
Remove `helpers` crate

loiclec:
# Pull Request

## What does this PR do?
Remove the `helpers` crates, because (I think) we don't use it. This should have been part of https://github.com/meilisearch/milli/pull/636 , but I forgot about it then :)


Kerollmops (approved):
Thank you!
bors merge

PR #642, opened 2022-09-21
Remove LTO in release profile

loiclec:
Since we can't enable it in Meilisearch (see https://github.com/meilisearch/meilisearch/pull/2717 ), we should not enable it in milli either. The goal is for milli's benchmarks to accurately represent its performance within meilisearch.


Kerollmops (approved):
Thank you, this makes sense!
bors merge

PR #662, opened 2022-10-12
Enhance word splitting strategy

### LINKED ISSUE #648: Enhance word splitting strategy
Today the word splitting strategy of the query tree is handled by the [function `split_best_frequency`](https://github.com/meilisearch/milli/blob/f8697075ea6b95b2d380e757af1724e75e0f21cf/milli/src/search/query_tree.rs#L266-L287).
This function split a word into two sub-words by looking at the frequency of the less frequent sub-word of each possible pair.

#### drawback
However, this frequency computation doesn't represent faithfully the frequency of the pair because these two sub-words can be considered individually frequent without being frequently near together in documents.

#### possible enhancement
Inspired by the [`index.rs#word_documents_count`](https://github.com/meilisearch/milli/blob/f8697075ea6b95b2d380e757af1724e75e0f21cf/milli/src/index.rs#L970-L976) method, a new method `word_pair_frequency` could be implemented in the trait [`search/query_tree.rs#Context`](https://github.com/meilisearch/milli/blob/f8697075ea6b95b2d380e757af1724e75e0f21cf/milli/src/search/query_tree.rs#L147-L159) using the [`word_pair_proximity_docids`](https://github.com/meilisearch/milli/blob/f8697075ea6b95b2d380e757af1724e75e0f21cf/milli/src/index.rs#L108-L109) database instead of the [`word_docids`](https://github.com/meilisearch/milli/blob/f8697075ea6b95b2d380e757af1724e75e0f21cf/milli/src/index.rs#L93-L94) one.

#### small warnings
- the trait `Context` is derived 2 times, one in [search/query_tree.rs#QueryTreeBuilder](https://github.com/meilisearch/milli/blob/f8697075ea6b95b2d380e757af1724e75e0f21cf/milli/src/search/query_tree.rs#L171) which is the real implementation, and one in [the tests](https://github.com/meilisearch/milli/blob/f8697075ea6b95b2d380e757af1724e75e0f21cf/milli/src/search/query_tree.rs#L822-L839)
- the database `word_pair_proximity_docids` has a different CODEC than `word_docids` but a length decoder is already implemented in [heed_codec/roaring_bitmap_length/](https://github.com/meilisearch/milli/tree/main/milli/src/heed_codec/roaring_bitmap_length)

#### Files expected to be modified
- [milli/src/search/query_tree.rs](https://github.com/meilisearch/milli/blob/main/milli/src/search/query_tree.rs)

akki1306:
# Pull Request

## Related issue
Fixes #648 

## What does this PR do?
- [split_best_frequency](https://github.com/meilisearch/milli/blob/55d889522b8b5b6e3dc6d5cf8d14a4f7f62c956b/milli/src/search/query_tree.rs#L282-L301) to use frequency of word pairs near together with proximity value of 1 instead of considering the frequency of individual words. Word pairs having max frequency are considered.

## PR checklist

### DIFF
--- a/milli/src/search/query_tree.rs
+++ b/milli/src/search/query_tree.rs
@@ -147,6 +147,12 @@ impl fmt::Debug for Query {
 trait Context {
     fn word_docids(&self, word: &str) -> heed::Result<Option<RoaringBitmap>>;
     fn synonyms<S: AsRef<str>>(&self, words: &[S]) -> heed::Result<Option<Vec<Vec<String>>>>;
+    fn word_documents_count(&self, word: &str) -> heed::Result<Option<u64>> {
+        match self.word_docids(word)? {
+            Some(rb) => Ok(Some(rb.len())),
+            None => Ok(None),
+        }
+    }
     /// Returns the minimum word len for 1 and 2 typos.
     fn min_word_len_for_typo(&self) -> heed::Result<(u8, u8)>;
     fn exact_words(&self) -> Option<&fst::Set<Cow<[u8]>>>;
@@ -276,7 +282,7 @@ impl<'a> QueryTreeBuilder<'a> {
 fn split_best_frequency<'a>(
     ctx: &impl Context,
     word: &'a str,
-) -> heed::Result<Option<(u64, &'a str, &'a str)>> {
+) -> heed::Result<Option<(&'a str, &'a str)>> {
     let chars = word.char_indices().skip(1);
     let mut best = None;
 
@@ -290,7 +296,7 @@ fn split_best_frequency<'a>(
         }
     }
 
-    Ok(best)
+    Ok(best.map(|(_, left, right)| (left, right)))
 }
 
 #[derive(Clone)]
@@ -359,7 +365,7 @@ fn create_query_tree(
             // 4. wrap all in an OR operation
             PrimitiveQueryPart::Word(word, prefix) => {
                 let mut children = synonyms(ctx, &[&word])?.unwrap_or_default();
-                if let Some((_, left, right)) = split_best_frequency(ctx, &word)? {
+                if let Some((left, right)) = split_best_frequency(ctx, &word)? {
                     children.push(Operation::Phrase(vec![left.to_string(), right.to_string()]));
                 }
                 let (word_len_one_typo, word_len_two_typo) = ctx.min_word_len_for_typo()?;
@@ -505,9 +511,7 @@ fn create_query_tree(
                 .filter(|(_, part)| !part.is_phrase())
                 .max_by_key(|(_, part)| match part {
                     PrimitiveQueryPart::Word(s, _) => {
-                        let (pair_freq, _, _) =
-                            split_best_frequency(ctx, s).unwrap_or_default().unwrap_or_default();
-                        pair_freq
+                        ctx.word_documents_count(s).unwrap_or_default().unwrap_or(u64::max_value());
                     }
                     _ => unreachable!(),
                 })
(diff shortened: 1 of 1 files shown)


milli/src/search/query_tree.rs
```diff
@@ -146,6 +146,12 @@ impl fmt::Debug for Query {
 
 trait Context {
     fn word_docids(&self, word: &str) -> heed::Result<Option<RoaringBitmap>>;
+    fn word_pair_proximity_docids(
+        &self,
+        right_word: &str,
+        left_word: &str,
+        proximity: u8,
+    ) -> heed::Result<Option<RoaringBitmap>>;
```
ManyTheFish, line 154: You should directly implement `word_pair_frequency` in the `TestContext` instead of creating this function.
```suggestion
```

milli/src/search/query_tree.rs
```diff
@@ -146,6 +146,12 @@ impl fmt::Debug for Query {
...
     fn synonyms<S: AsRef<str>>(&self, words: &[S]) -> heed::Result<Option<Vec<Vec<String>>>>;
     fn word_documents_count(&self, word: &str) -> heed::Result<Option<u64>> {
```
ManyTheFish, line 156: I think this function is no more needed.

ManyTheFish (changes requested):
Hey @akki1306,
I requested some small changes, however, the general implementation is good.
I let you make the changes and I'll re-review after. 👍

> Don't forget to ensure that all the tests are successful by running `cargo test`.

milli/src/search/query_tree.rs
akki1306: Removed as recommended

milli/src/search/query_tree.rs
```diff
@@ -498,7 +510,9 @@ fn create_query_tree(
                 .filter(|(_, part)| !part.is_phrase())
                 .max_by_key(|(_, part)| match part {
                     PrimitiveQueryPart::Word(s, _) => {
-                        ctx.word_documents_count(s).unwrap_or_default().unwrap_or(u64::max_value())
+                        let (pair_freq, _, _) =
+                            split_best_frequency(ctx, s).unwrap_or_default().unwrap_or_default();
+                        pair_freq
```
ManyTheFish, line 510: Oh! Sorry, I forgot this part of the code 🙏.
Could you come back to the previous version, please?

akki1306:
Hi @ManyTheFish, I have addressed the comments. Please take a look.

ManyTheFish (changes requested):
re @akki1306, I requested some changes due to an oversight on my side.

milli/src/search/query_tree.rs
akki1306: Sure, I will add back word_documents_count 👍 and revert this part

akki1306: Done, PTAL.

akki1306:
Thanks @ManyTheFish, I was trying to understand the code regarding the single use case for the word_documents_count method. What I understood is while resolving the primitive query part we are considering word pairs with the highest frequency whereas if the TermMatchingStrategy is set to Frequency, we use word_documents_count. Please let me know if the above understanding is correct or if you could throw some light in this direction, was curious to know about the construction of the query tree.

ManyTheFish:
@akki1306, not really.
Sometimes, an End User skips a whitespace while typing his search query, it's probably an oversight.
To catch this oversight and avoid the End User rewriting his query, we try to split each searched word based on the subwords frequency. For instance, in the word `bypassword`, the EndUser forgot a space, but where?
The function you enhanced has this Job, split a word into 2 in order to find the most relevant result. More precisely,
the previous version was searching the frequency of the 2 subwords independently, let's imagine that `bypass` and `word` were both more frequent than `by` and `password` but the frequency of seeing near together is really low, the previous version would choose `bypass` and `word` despite the chance to find anything relevant with this split.
Your modification takes now into account the frequency of the word near together.

The TermMatchingStrategy is another story, it defines the rule: "In which order do we ignore query words to retrieve more documents?":
- from the last typed word to the first?
- from the first typed one to the last?
- from the most frequent word to the least?
- from the shortest word to the longest?

In the third case, the interesting frequency is the frequency of the word alone and not near another word.

akki1306:
Thanks for the detailed explanation @ManyTheFish, it really helped improve my understanding of the query tree construction functionality 👍

PR #665, opened 2022-10-13
Fixing piles of clippy errors.

ehiggs:
## Related issue
No issue fixed. Simply cleaning up some code for clippy on the march towards a clean build when #659 is merged.

## What does this PR do?
Most of these are calling clone when the struct supports Copy.

Many are using & and &mut on `self` when the function they are called from already has an immutable or mutable borrow so this isn't needed.

I tried to stay away from actual changes or places where I'd have to name fresh variables.

## PR checklist
Please check if your PR fulfills the following requirements:
- [x] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [x] Have you read the contributing guidelines?
- [x] Have you made sure that the title is accurate and descriptive of the changes?

PR #659, opened 2022-10-10
Fix clippy error to add clippy job on Ci

unvalley:
## Related PR
This PR is for #673 

## What does this PR do?
- ~~add `Run Clippy` job to CI (rust.yml)~~
- apply `cargo clippy --fix` command
- fix some `cargo clippy` error manually (but warnings still remain on tests)

## PR checklist
Please check if your PR fulfills the following requirements:
- [x] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [x] Have you read the contributing guidelines?
- [x] Have you made sure that the title is accurate and descriptive of the changes?

PR #668, opened 2022-10-24
Fix many Clippy errors part 2

ehiggs:
This brings us a step closer to enforcing clippy on each build.

# Pull Request

## Related issue
This does not fix any issue outright, but it is a second round of fixes for clippy after https://github.com/meilisearch/milli/pull/665. This should contribute to fixing https://github.com/meilisearch/milli/pull/659.

## What does this PR do?

Satisfies many issues for clippy. The complaints are mostly:

* Passing reference where a variable is already a reference.
* Using clone where a struct already implements `Copy`
* Using `ok_or_else` when it is a closure that returns a value instead of using the closure to call function (hence we use `ok_or`)
* Unambiguous lifetimes don't need names, so we can just use `'_`
* Using `return` when it is not needed as we are on the last expression of a function.

## PR checklist

### DIFF
--- a/milli/src/asc_desc.rs
+++ b/milli/src/asc_desc.rs
@@ -70,7 +70,7 @@ impl FromStr for Member {
     type Err = AscDescError;
 
     fn from_str(text: &str) -> Result<Member, Self::Err> {
-        match text.strip_prefix("_geoPoint(").and_then(|text| text.strip_suffix(")")) {
+        match text.strip_prefix("_geoPoint(").and_then(|text| text.strip_suffix(')')) {
             Some(point) => {
                 let (lat, lng) = point
                     .split_once(',')
--- a/milli/src/fields_ids_map.rs
+++ b/milli/src/fields_ids_map.rs
@@ -65,7 +65,7 @@ impl FieldsIdsMap {
     }
 
     /// Iterate over the ids in the order of the ids.
-    pub fn ids<'a>(&'a self) -> impl Iterator<Item = FieldId> + 'a {
+    pub fn ids(&'_ self) -> impl Iterator<Item = FieldId> + '_ {
         self.ids_names.keys().copied()
     }
 
--- a/milli/src/search/query_tree.rs
+++ b/milli/src/search/query_tree.rs
@@ -313,7 +313,7 @@ pub struct TypoConfig<'a> {
 
 /// Return the `QueryKind` of a word depending on `authorize_typos`
 /// and the provided word length.
-fn typos(word: String, authorize_typos: bool, config: TypoConfig<'_>) -> QueryKind {
+fn typos(word: String, authorize_typos: bool, config: TypoConfig) -> QueryKind {
     if authorize_typos && !config.exact_words.map_or(false, |s| s.contains(&word)) {
         let count = word.chars().count().min(u8::MAX as usize) as u8;
         if count < config.word_len_one_typo {
--- a/milli/src/search/query_tree.rs
+++ b/milli/src/search/query_tree.rs
@@ -188,8 +188,8 @@ impl<'a> Context for QueryTreeBuilder<'a> {
     }
 
     fn min_word_len_for_typo(&self) -> heed::Result<(u8, u8)> {
-        let one = self.index.min_word_len_one_typo(&self.rtxn)?;
-        let two = self.index.min_word_len_two_typos(&self.rtxn)?;
+        let one = self.index.min_word_len_one_typo(self.rtxn)?;
+        let two = self.index.min_word_len_two_typos(self.rtxn)?;
         Ok((one, two))
     }
 
@@ -207,7 +207,7 @@ impl<'a> Context for QueryTreeBuilder<'a> {
         self.index
             .word_pair_proximity_docids
             .remap_data_type::<CboRoaringBitmapLenCodec>()
-            .get(&self.rtxn, &key)
+            .get(self.rtxn, &key)
     }
 }
 
@@ -313,7 +313,7 @@ pub struct TypoConfig<'a> {
 
 /// Return the `QueryKind` of a word depending on `authorize_typos`
 /// and the provided word length.
-fn typos<'a>(word: String, authorize_typos: bool, config: TypoConfig<'a>) -> QueryKind {
+fn typos(word: String, authorize_typos: bool, config: TypoConfig<'_>) -> QueryKind {
     if authorize_typos && !config.exact_words.map_or(false, |s| s.contains(&word)) {
         let count = word.chars().count().min(u8::MAX as usize) as u8;
         if count < config.word_len_one_typo {
@@ -556,7 +556,7 @@ fn create_matching_words(
                     for synonym in synonyms {
                         let synonym = synonym
                             .into_iter()
-                            .map(|syn| MatchingWord::new(syn.to_string(), 0, false))
+                            .map(|syn| MatchingWord::new(syn, 0, false))
                             .collect();
                         matching_words.push((synonym, vec![id]));
                     }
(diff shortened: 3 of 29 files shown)


milli/src/asc_desc.rs
```diff
@@ -70,7 +70,7 @@ impl FromStr for Member {
     type Err = AscDescError;
 
     fn from_str(text: &str) -> Result<Member, Self::Err> {
-        match text.strip_prefix("_geoPoint(").and_then(|text| text.strip_suffix(")")) {
+        match text.strip_prefix("_geoPoint(").and_then(|text| text.strip_suffix(')')) {
```
ehiggs, line 73: This is one I wasn't 100% sure of as we are changing from `[')', '\0']` to `')'` I think.

ehiggs:
The remaining clippy errors (not in this PR) seem like they will need configuration or some abstractions.

1. Functions with too many parameters. Should probably be silenced.
2. Types that are too complicated can be named. This could be easily done by making a `type` declaration before use, but the essence of what clippy is telling us is to refactor so the types have a more robust meaning. So abbreviating the names in a name doesn't do this justice. Can be silenced.
3. [Creating an empty range](https://rust-lang.github.io/rust-clippy/master/index.html#reversed_empty_ranges) (`1..=0`) (`milli/src/update/available_documents_ids.rs:24:29`) this seems intentional in the code and I looked into the docs to see if there is a more explicit solution but I didn't see one. Silence?
4. [Something about `concat_arrays!` and Drop that I didn't fully understand in this context](https://rust-lang.github.io/rust-clippy/master/index.html#drop_non_drop)
5. [Two instances of iter -> collect into Vec -> return iter](https://rust-lang.github.io/rust-clippy/master/index.html#needless_collect) in `milli/src/search/criteria/asc_desc.rs`. I tried to fix this but the functions rely on data in the functions so it's collected to make a copy. I didn't see a way past this without making copies.

milli/src/fields_ids_map.rs
```diff
@@ -65,7 +65,7 @@ impl FieldsIdsMap {
     }
 
     /// Iterate over the ids in the order of the ids.
-    pub fn ids<'a>(&'a self) -> impl Iterator<Item = FieldId> + 'a {
+    pub fn ids(&'_ self) -> impl Iterator<Item = FieldId> + '_ {
```
ManyTheFish, line 68: Is putting an anonymous lifetime necessary here?
```suggestion
    pub fn ids(&self) -> impl Iterator<Item = FieldId> {
```

milli/src/search/query_tree.rs
```diff
@@ -313,7 +313,7 @@ pub struct TypoConfig<'a> {
 
 /// Return the `QueryKind` of a word depending on `authorize_typos`
 /// and the provided word length.
-fn typos<'a>(word: String, authorize_typos: bool, config: TypoConfig<'a>) -> QueryKind {
+fn typos(word: String, authorize_typos: bool, config: TypoConfig<'_>) -> QueryKind {
```
ManyTheFish, line 316: Is this anonymous lifetime necessary?
```suggestion
fn typos(word: String, authorize_typos: bool, config: TypoConfig) -> QueryKind {
```

milli/src/update/facets.rs
```diff
@@ -262,8 +262,8 @@ impl<'t, 'u, 'i> Facets<'t, 'u, 'i> {
 /// 1. a vector of grenad::Reader. The reader at index `i` corresponds to the elements of level `i + 1`
 /// that must be inserted into the database.
 /// 2. a roaring bitmap of all the document ids present in the database
-fn compute_facet_number_levels<'t>(
-    rtxn: &'t heed::RoTxn,
+fn compute_facet_number_levels(
+    rtxn: &'_ heed::RoTxn,
```
ManyTheFish, line 266: Is this anonymous lifetime necessary?
```suggestion
    rtxn: &heed::RoTxn,
```

ManyTheFish (changes requested):

ManyTheFish:
Hello @ehiggs, thank you for your contribution,
I put below the answers to your questions.

> The remaining clippy errors (not in this PR) seem like they will need configuration or some abstractions.
> 
>     1. Functions with too many parameters. Should probably be silenced.

You may be right on this, it would avoid the next Clippy contribution to invest time in these functions

>     2. Types that are too complicated can be named. This could be easily done by making a `type` declaration before use, but the essence of what clippy is telling us is to refactor so the types have a more robust meaning. So abbreviating the names in a name doesn't do this justice. Can be silenced.

Let's skip this one, Clippy is right, we should wrap complex type declarations under an understandable name, but it's a choice to do internally, and sometime it will need some refactoring.

>     3. [Creating an empty range](https://rust-lang.github.io/rust-clippy/master/index.html#reversed_empty_ranges) (`1..=0`) (`milli/src/update/available_documents_ids.rs:24:29`) this seems intentional in the code and I looked into the docs to see if there is a more explicit solution but I didn't see one. Silence?

It's completely intentional, there are other solutions, but I think it would need the function to return a boxed dynamic iterator. 🤔 And we would like to avoid this.

>     4. [Something about `concat_arrays!` and Drop that I didn't fully understand in this context](https://rust-lang.github.io/rust-clippy/master/index.html#drop_non_drop)

which one?

>     5. [Two instances of iter -> collect into Vec -> return iter](https://rust-lang.github.io/rust-clippy/master/index.html#needless_collect) in `milli/src/search/criteria/asc_desc.rs`. I tried to fix this but the functions rely on data in the functions so it's collected to make a copy. I didn't see a way past this without making copies.

The related issue in itertool is still open. So to fix this one, the only solution I see is to recode your own iterator to resolve the lifetime issue, but, I don't think we would gain anything by doing it, so I suggest ignoring it. 🤔

milli/src/fields_ids_map.rs
ehiggs: This one is needed:

```
hidden type for `impl Iterator<Item = u16>` captures lifetime that does not appear in boundsrustc[E0700](https://doc.rust-lang.org/error-index.html#E0700)
fields_ids_map.rs(68, 16): hidden type `std::iter::Copied<std::collections::btree_map::Keys<'_, u16, std::string::String>>` captures the anonymous lifetime defined here
fields_ids_map.rs(68, 55): to declare that the `impl Trait` captures `'_`, you can add an explicit `'_` lifetime bound: ` + '_`
```

milli/src/search/query_tree.rs
ehiggs: This one is not needed :tada:

milli/src/update/facets.rs
ehiggs: This one is not needed.

milli/src/search/query_tree.rs
ehiggs: fixed

milli/src/update/facets.rs
ehiggs: fixed

ehiggs:
> You may be right on this, it would avoid the next Clippy contribution to invest time in these functions

Shall I put a decl to silence this?

ManyTheFish:
> Shall I put a decl to silence this in this PR?

Yes you can, but do it in a separate commit in case of this change is too ugly

> Sorry to be a nuisance but do you mean "let's skip this one" to mean let's silence it? Or let's keep it flagging the code and take some time to refactor it.

Don't do any changes to this, this should be changed during a real refactor, which is not the purpose of your PR. So let's keep it flagging the code.

>  ... concat_arrays!

OK! Well, I don't know and I find this implementation really clear. So I would prefer to Silence it.

ehiggs:
In the last commit:

1. I silenced the 'too_many_arguments'  based on discussion above.
2. I silenced the issue with `1..=0` (which is not allowed by default clippy).

With this in place, clippy can run with a 0 exit code despite printing warnings on various issues. This means we can enable clippy in CI under #659.

ManyTheFish (approved):
Hey @ehiggs! Thank you for your contribution, it looks good to me!
I let bors run all the tests, if everything goes well, your PR will be automatically merged.

Bors merge

PR #689, opened 2022-11-08
Handle non-finite floats consistently in filters

dureuill:
# Pull Request

## Related issue

Related meilisearch/meilisearch#3000

### User

- Filters using `field = inf`, (or `infinite`, `NaN`) now match the value as a string rather than returning an internal error.
- Filters using `field < inf` (or other comparison operators) now return an invalid_filter error rather than returning an internal error, much like when using `field < aaa`.

### Implementation

- Add new `NonFiniteFloat` error variants to the filter-parser errors
- Add `Token::parse_as_finite_float` that can fail both when the string is not a float and when the float is not finite
- Refactor `Filter::inner_evaluate` to always use `parse_as_finite_float` instead of just `parse`
- Add corresponding tests

## PR checklist

### DIFF
--- a/milli/src/search/facet/filter.rs
+++ b/milli/src/search/facet/filter.rs
@@ -698,4 +698,60 @@ mod tests {
         let option = Filter::from_str("     ").unwrap();
         assert_eq!(option, None);
     }
+
+    #[test]
+    fn non_finite_float() {
+        let index = TempIndex::new();
+
+        index
+            .update_settings(|settings| {
+                settings.set_searchable_fields(vec![S("price")]); // to keep the fields order
+                settings.set_filterable_fields(hashset! { S("price") });
+            })
+            .unwrap();
+        index
+            .add_documents(documents!([
+                {
+                    "id": "test_1",
+                    "price": "inf"
+                },
+                {
+                    "id": "test_2",
+                    "price": "2000"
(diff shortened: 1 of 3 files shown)


milli/src/search/facet/filter.rs
```diff
@@ -689,4 +698,61 @@ mod tests {
...
+        let filter = Filter::from_str("price = NaN").unwrap().unwrap();
+        let result = filter.evaluate(&rtxn, &index).unwrap();
+        assert!(result.is_empty());
+        let filter = Filter::from_str("price < NaN").unwrap().unwrap();
+        assert!(matches!(
+            filter.evaluate(&rtxn, &index),
+            Err(crate::Error::UserError(crate::error::UserError::InvalidFilter(_)))
+        ));
+
+        let rtxn = index.read_txn().unwrap();
```
irevoire, line 748:
```suggestion
```

irevoire (changes requested):
Nice, thank you.
@loiclec if it’s ok for you as well you can merge

loiclec:
Yes, seems perfect :)

Kerollmops:
Could you please check the categories of floats that we accept when we index? Do we also only accept `.is_finite()` numbers?

PR #690, opened 2022-11-08
Fix soft deleted bug settings

Kerollmops:
This PR fixes a bug that was triggered by reindexing the documents when changing the settings, we were indexing documents that were soft-deleted instead of ignoring them.

Related to https://github.com/meilisearch/meilisearch/issues/3021 and https://github.com/meilisearch/meilisearch/issues/2945.

### DIFF
--- a/milli/src/update/index_documents/transform.rs
+++ b/milli/src/update/index_documents/transform.rs
@@ -578,9 +578,8 @@ impl<'a, 'i> Transform<'a, 'i> {
         );
 
         let mut obkv_buffer = Vec::new();
-        for result in self.index.documents.iter(wtxn)? {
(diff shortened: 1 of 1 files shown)


milli/src/update/settings.rs
```diff
@@ -1494,4 +1495,34 @@ mod tests {
             })
             .unwrap();
     }
+
+    #[test]
+    fn settings_mut_ignore_soft_deleted() {
```
ManyTheFish, line 1500:
```suggestion
    fn settings_must_ignore_soft_deleted() {
```

ManyTheFish (changes requested):

ManyTheFish (approved):
bors merge

PR #706, opened 2022-11-23
Limit the reindexing caused by updating settings when not needed

GregoryConrad:
## What does this PR do?
When updating index settings using `update::Settings`, sometimes a `reindex` of `update::Settings` is triggered when it doesn't need to be. This PR aims to prevent those unnecessary `reindex` calls.

For reference, here is a snippet from the current `execute` method in `update::Settings`:
```rust
// ...
if stop_words_updated
    || faceted_updated
    || synonyms_updated
    || searchable_updated
    || exact_attributes_updated
{
    self.reindex(&progress_callback, &should_abort, old_fields_ids_map)?;
}
```

- [x] `faceted_updated` - looks good as-is ✅
- [x] `stop_words_updated` - looks good as-is ✅
- [x] `synonyms_updated` - looks good as-is ✅
- [x] `searchable_updated` - fixed in this PR
- [x] `exact_attributes_updated` - fixed in this PR

## PR checklist

### DIFF
--- a/milli/src/update/settings.rs
+++ b/milli/src/update/settings.rs
@@ -349,12 +349,17 @@ impl<'a, 't, 'u, 'i> Settings<'a, 't, 'u, 'i> {
     fn update_searchable(&mut self) -> Result<bool> {
         match self.searchable_fields {
             Setting::Set(ref fields) => {
-                let did_change = self
-                    .index
-                    .searchable_fields(self.wtxn)?
-                    .map(|f| f.into_iter().map(String::from).collect::<Vec<_>>())
-                    .map(|old_fields| fields != &old_fields)
-                    .unwrap_or(true); // if old_fields was None before, it was changed
+                // Check to see if the searchable fields changed before doing anything else
+                let old_fields = self.index.searchable_fields(self.wtxn)?;
+                let did_change = match old_fields {
+                    // If old_fields is Some, let's check to see if the fields actually changed
+                    Some(old_fields) => {
+                        let new_fields = fields.iter().map(String::as_str).collect::<Vec<_>>();
+                        new_fields != old_fields
+                    }
+                    // If old_fields is None, the fields have changed (because they are being set)
+                    None => true,
+                };
                 if !did_change {
                     return Ok(false);
                 }
--- a/milli/src/update/settings.rs
+++ b/milli/src/update/settings.rs
@@ -465,14 +465,23 @@ impl<'a, 't, 'u, 'i> Settings<'a, 't, 'u, 'i> {
     fn update_exact_attributes(&mut self) -> Result<bool> {
         match self.exact_attributes {
             Setting::Set(ref attrs) => {
-                let attrs = attrs.iter().map(String::as_str).collect::<Vec<_>>();
-                self.index.put_exact_attributes(self.wtxn, &attrs)?;
-                Ok(true)
+                let old_attrs = self
+                    .index
+                    .exact_attributes(self.wtxn)?
+                    .iter()
+                    .cloned()
+                    .map(String::from)
+                    .collect::<HashSet<String>>();
+
+                if attrs != &old_attrs {
+                    let attrs = attrs.iter().map(String::as_str).collect::<Vec<_>>();
+                    self.index.put_exact_attributes(self.wtxn, &attrs)?;
+                    Ok(true)
+                } else {
+                    Ok(false)
+                }
             }
-            Setting::Reset => {
-                self.index.delete_exact_attributes(self.wtxn)?;
-                Ok(true)
-            }
+            Setting::Reset => Ok(self.index.delete_exact_attributes(self.wtxn)?),
             Setting::NotSet => Ok(false),
         }
     }
--- a/milli/src/update/settings.rs
+++ b/milli/src/update/settings.rs
@@ -478,13 +478,8 @@ impl<'a, 't, 'u, 'i> Settings<'a, 't, 'u, 'i> {
     fn update_exact_attributes(&mut self) -> Result<bool> {
         match self.exact_attributes {
             Setting::Set(ref attrs) => {
-                let old_attrs = self
-                    .index
-                    .exact_attributes(self.wtxn)?
-                    .iter()
-                    .cloned()
-                    .map(String::from)
-                    .collect::<HashSet<String>>();
+                let old_attrs = self.index.exact_attributes(self.wtxn)?;
+                let old_attrs = old_attrs.into_iter().map(String::from).collect::<HashSet<_>>();
 
                 if attrs != &old_attrs {
                     let attrs = attrs.iter().map(String::as_str).collect::<Vec<_>>();
--- a/milli/src/update/settings.rs
+++ b/milli/src/update/settings.rs
@@ -373,13 +373,11 @@ impl<'a, 't, 'u, 'i> Settings<'a, 't, 'u, 'i> {
                     &new_fields_ids_map,
                 )?;
                 self.index.put_fields_ids_map(self.wtxn, &new_fields_ids_map)?;
+                Ok(true)
             }
-            Setting::Reset => {
-                self.index.delete_all_searchable_fields(self.wtxn)?;
-            }
+            Setting::Reset => Ok(self.index.delete_all_searchable_fields(self.wtxn)?),
             Setting::NotSet => return Ok(false),
         }
-        Ok(true)
     }
 
     fn update_stop_words(&mut self) -> Result<bool> {
--- a/milli/src/update/settings.rs
+++ b/milli/src/update/settings.rs
@@ -349,6 +349,16 @@ impl<'a, 't, 'u, 'i> Settings<'a, 't, 'u, 'i> {
     fn update_searchable(&mut self) -> Result<bool> {
         match self.searchable_fields {
             Setting::Set(ref fields) => {
+                let did_change = self
+                    .index
+                    .searchable_fields(self.wtxn)?
+                    .map(|f| f.into_iter().map(String::from).collect::<Vec<_>>())
+                    .map(|old_fields| fields != &old_fields)
+                    .unwrap_or(true); // if old_fields was None before, it was changed
+                if !did_change {
+                    return Ok(false);
+                }
+
                 // every time the searchable attributes are updated, we need to update the
                 // ids for any settings that uses the facets. (distinct_fields, filterable_fields).
                 let old_fields_ids_map = self.index.fields_ids_map(self.wtxn)?;
@@ -376,7 +386,7 @@ impl<'a, 't, 'u, 'i> Settings<'a, 't, 'u, 'i> {
                 Ok(true)
             }
             Setting::Reset => Ok(self.index.delete_all_searchable_fields(self.wtxn)?),
-            Setting::NotSet => return Ok(false),
+            Setting::NotSet => Ok(false),
         }
     }
 
--- a/milli/src/index.rs
+++ b/milli/src/index.rs
@@ -1145,8 +1145,8 @@ impl Index {
     }
 
     /// Clears the exact attributes from the store.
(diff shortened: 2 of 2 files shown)


GregoryConrad:
For my logic in `update_searchable`'s `Setting::Reset` match branch to work, I assume `index::delete_all_searchable_fields` returns true when reindexing needs to happen. I *am not sure* if that assumption is correct; thus, please verify that the below should return true when reindexing needs to happen:
```rust
pub(crate) fn delete_all_searchable_fields(&self, wtxn: &mut RwTxn) -> heed::Result<bool> {
    self.delete_searchable_fields(wtxn)?;
    self.delete_user_defined_searchable_fields(wtxn)
}
```
To clarify, I am not sure if this should actually return (for example):
```rust
pub(crate) fn delete_all_searchable_fields(&self, wtxn: &mut RwTxn) -> heed::Result<bool> {
    self.delete_searchable_fields(wtxn)? && self.delete_user_defined_searchable_fields(wtxn)?
}
```

milli/src/update/settings.rs
```diff
@@ -349,6 +349,16 @@ impl<'a, 't, 'u, 'i> Settings<'a, 't, 'u, 'i> {
...
             Setting::Set(ref fields) => {
+                let did_change = self
+                    .index
+                    .searchable_fields(self.wtxn)?
+                    .map(|f| f.into_iter().map(String::from).collect::<Vec<_>>())
+                    .map(|old_fields| fields != &old_fields)
+                    .unwrap_or(true); // if old_fields was None before, it was changed
+                if !did_change {
+                    return Ok(false);
+                }
```
Kerollmops, line 360: Can you confirm that you are retrieving the previous searchable to compare them with the new ones and if nothing changes you return false, right? It is a typical situation when users send the whole set of settings and don't care of the previous ones...

milli/src/update/settings.rs
```diff
@@ -465,14 +473,23 @@ impl<'a, 't, 'u, 'i> Settings<'a, 't, 'u, 'i> {
...
+                    .map(String::from)
+                    .collect::<HashSet<String>>();
+
+                if attrs != &old_attrs {
+                    let attrs = attrs.iter().map(String::as_str).collect::<Vec<_>>();
+                    self.index.put_exact_attributes(self.wtxn, &attrs)?;
+                    Ok(true)
+                } else {
+                    Ok(false)
+                }
```
Kerollmops, line 490: And you are doing exactly the same here, right?

Kerollmops (commented):
Thank you very much for your changes, it looks amazing indeed, it will speedup a lot of workflows where users were always sending the searchable attributes ❤️

milli/src/update/settings.rs
GregoryConrad: That is correct! Would you like me to break apart that chained logic to make it more readable?

milli/src/update/settings.rs
GregoryConrad: Correct. Here I compare the two `HashSet`s of old/new, whereas up above I compared the two `Vec`s of old/new.

Also, if you would mind taking a look at [this comment above](https://github.com/meilisearch/milli/pull/706#issuecomment-1325885217), I would really appreciate it. Not sure if that assumption I made for the `Settings::Reset` case is correct.

GregoryConrad:
^^ Don't know why GitHub removed that request for review from @ManyTheFish. I only requested a re-review from @Kerollmops. Sorry about that

milli/src/update/settings.rs
Kerollmops: Indeed, it would probably make this a lot more readable 😅

Kerollmops:
> For my logic in `update_searchable`'s `Setting::Reset` match branch to work, I assume `index::delete_all_searchable_fields` returns true when reindexing needs to happen. I _am not sure_ if that assumption is correct; thus, please verify that the below should return true when reindexing needs to happen:
> 
> ```rust
> pub(crate) fn delete_all_searchable_fields(&self, wtxn: &mut RwTxn) -> heed::Result<bool> {
>     self.delete_searchable_fields(wtxn)?;
>     self.delete_user_defined_searchable_fields(wtxn)
> }
> ```
> 
> To clarify, I am not sure if this should actually return (for example):
> 
> ```rust
> pub(crate) fn delete_all_searchable_fields(&self, wtxn: &mut RwTxn) -> heed::Result<bool> {
>     Ok(self.delete_searchable_fields(wtxn)? && self.delete_user_defined_searchable_fields(wtxn)?)
> }
> ```

The delete functions should be mapped on [the LMDB `mdb_del` function](http://www.lmdb.tech/doc/group__mdb.html#gab8182f9360ea69ac0afd4a4eaab1ddb0) and return `true` if the value was found in the database. However, ensure you don't use the latest code you published, as you'll skip the right part of the `&&` if the left part returns `false`.

So the short answer is: yes, the *delete_** methods must return `true` if the value existed and was effectively deleted.

milli/src/update/settings.rs
GregoryConrad: Thanks for the answer (and correcting that example I gave--I overlooked the `&&` behavior); however, I think I wasn't specific enough:

Currently, `delete_all_searchable_fields`:
- returns the result of executing `self.delete_user_defined_searchable_fields(wtxn)`
- ignores the result of executing `self.delete_searchable_fields(wtxn)?;`

When should reindexing happen (logically speaking here)? Here are a few options that I thought of:
- Whenever `delete_searchable_fields` returns `true`?
- when `delete_user_defined_searchable_fields` and `delete_searchable_fields` both return true?
- when at least one of the above return true?

I am not versed well-enough in milli's inner-workings to be able to answer this question myself, as I am not 100% clear how `user_defined_searchable_fields` and `searchable_fields` relate to reindexing.

Kerollmops: I would say _when at least one of the above returns `true`_. Thank you very much for your work on the engine BTW, very pleasant to see you involved :)

GregoryConrad: This is in relation to optimizing the `Settings::Reset` behavior--if a reset was selected, but no reset is needed, then we shouldn't need to reindex. However, we also have to be careful to not miss a reindex with `Settings::Reset` when we actually need one. We could be safe and return an `||` of the results of the two above methods, but this could cause reindexes when they aren't needed.

GregoryConrad: > I would say _when at least one of the above returns `true`_. Thank you very much for your work on the engine BTW, very pleasant to see you involved :)

@Kerollmops Thanks! I'll apply the fix for that in a couple hours.

GregoryConrad: I made the appropriate change in 87e2bc3.

@irevoire I am roping you in here as an extra pair of eyes because of your work in f586028f9a8250e462c43484c6ce4867dc2d5aff 6 months ago. Does my new implementation of `delete_all_searchable_fields` in 87e2bc3 return true *only* when reindexing needs to occur? I am trying hard not to mess this up because I don't want to cause reindexing when it is actually not needed.

If I have it correct in 87e2bc3, maybe we could also add a check in to see if the user had defined any fields previously and only call `delete_searchable_fields`/`delete_all_searchable_fields` when the user had previously defined searchable fields?

irevoire: Hey @GregoryConrad,
> Does my new implementation of delete_all_searchable_fields in https://github.com/meilisearch/milli/commit/87e2bc3beda573a72c87647d614393d524bd5b44 return true only when reindexing needs to occur?

Yep, I think it should be ok like that.

> If I have it correct in https://github.com/meilisearch/milli/commit/87e2bc3beda573a72c87647d614393d524bd5b44, maybe we could also add a check in to see if the user had defined any fields previously and only call `delete_searchable_fields`/`delete_all_searchable_fields` when the user had previously defined searchable fields?

I'm not sure I see the point in doing that. If the user had no searchable_field defined, then calling `delete_searchable_fields` or `delete_all_searchable_fields` will return `false` and won't trigger a re-indexing, right?

Kerollmops (approved):
What you've done looks good to me. I'll wait for you to answer @irevoire and would like to know what @irevoire thinks of merging this now or waiting for @loiclec to fix the issue of the duplicated and _Missing key in the documents database_ bug. It would be safer to wait for the fix, but I am also quite happy if we could provide that to the users in v0.30.1.

irevoire (approved):
Yep, it looks good to me as well.

I'm not sure we want to push this change in the v0.30.1 since it's not a bug fix, though :scream: 
What do you think @curquiza?
It looks pretty safe to me, but it's typically the kind of thing we said we would not do :joy:

curquiza:
Discussed internally [here](https://meilisearch.slack.com/archives/G01A1F4KVGU/p1669739900852469): #707 is safe to be integrated into v0.30.1.
So don't merge this PR until v0.30.1 is released. I pass it as a draft to avoid any issues!

milli/src/update/settings.rs
GregoryConrad: > If the user had no searchable_field defined, then calling delete_searchable_fields or delete_all_searchable_fields will return false and won't trigger a re-indexing, right?

I wasn't sure if `delete_searchable_fields` *actually* would return false in that case, which is why I brought it up. Reason: it seemed like `searchable_fields` references the currently searchable fields, which will always exist, and wiping those out would need a reindex (even if you just populate the same searchable fields again in the index). Thus, the idea for the optimization. I can experiment with the current behavior in a little bit and see for sure how the index acts

GregoryConrad:
@curquiza Thanks for the update! Just as a heads up, please hold off on merging this until I get a chance to address [this](https://github.com/meilisearch/milli/pull/706#discussion_r1035219509). I need to experiment with the index's current behavior to see if I might be able to throw in another optimization (or if it is good as-is).

milli/src/update/settings.rs
GregoryConrad: Looks like we're all set as-is:
```rust
        println!("START");
        let index = TempIndex::new();

        index
            .add_documents(documents!([
                { "id": 1, "name": "kevin", "age": 23 },
                { "id": 2, "name": "kevina", "age": 21},
                { "id": 3, "name": "benoit", "age": 34 }
            ]))
            .unwrap();

        index.update_settings(|settings| {
            settings.set_searchable_fields(vec!("name".to_string()));
        }).unwrap();
        index.update_settings(|settings| {
            settings.reset_searchable_fields();
        }).unwrap();
        index.update_settings(|settings| {
            settings.reset_searchable_fields();
        }).unwrap();

        println!("FINISH");
```
Yields the following output:
```
START
REINDEXING!
REINDEXING!
FINISH
```
Which is expected (2 reindexes). I was unsure if the second `reset_searchable_fields` call would cause an extra reindex in the implementation as-is.

irevoire: Yep, I think this process is now totally _clear_ to you! :grin: 

Since you went pretty far in understanding how it works, it doesn't impact your PR, but maybe you would be interested in knowing how the « real » searchable fields work compared to the user-defined ones.

Basically, IIRC, if you set `doggo` as a searchable field and have this document:
```JSON
{
  "id": 1,
  "doggo": {
    "name": "bob",
    "race": "Bernese mountain dog"
  }
}
```

Then your user-defined searchable fields are going to be `doggo`, while the real searchable fields are going to be `doggo.name, doggo.race` (and maybe even `doggo`, I don't remember).
And this second list might update itself while you add more documents.
If you add a second document like that:
```JSON
{
  "id": 2,
  "doggo": {
    "name": "jean",
    "race": "golden retriever",
    "comment": "Jean is an adorable golden retriever that loves to be cuddled."
  }
}
```

Then while your user-defined searchable fields still won't change, your real/hidden searchable fields will add a third `doggo.comment` field.

GregoryConrad: Thanks for that amazing explanation @irevoire !

curquiza:
We can merge this PR to main finally, because other PRs have been merged into `main`
We will manage with git branch to do a custom release with only the needed commit

bors merge

PR #707, opened 2022-11-24
Add all_obkv_to_json function

GregoryConrad:
## What does this PR do?
When embedding milli in an application (other than Meilisearch), it often makes sense to not use the `displayed_attributes` functionality and instead just use milli as a full document store. Thus, this PR adds a function, `all_obkv_to_json`, to supplement the already exposed `milli::obkv_to_json` so that those embedding milli *do not* need to deal with `displayed_attributes` if they don't need to.

~This PR also introduces a slight breaking change: `obkv_to_json` now accepts a reference to `obkv::KvReaderU16` instead of taking ownership of it. As far as I can tell, this seems like a change for the better (`obkv_to_json` only acts upon `obkv` rather than consuming it), but I can change it back if you so desire.~ (reverted in [935a724](https://github.com/meilisearch/milli/pull/707/commits/935a724c570b3b851da544b96734ce20cf280f2c))

## PR checklist

### DIFF
--- a/milli/src/lib.rs
+++ b/milli/src/lib.rs
@@ -88,7 +88,7 @@ pub fn absolute_from_relative_position(field_id: FieldId, relative: RelativePosi
 pub fn obkv_to_json(
     displayed_fields: &[FieldId],
     fields_ids_map: &FieldsIdsMap,
-    obkv: obkv::KvReaderU16,
+    obkv: &obkv::KvReaderU16,
 ) -> Result<Object> {
     displayed_fields
         .iter()
@@ -103,6 +103,12 @@ pub fn obkv_to_json(
             Ok((name.to_owned(), value))
         })
         .collect()
+}
+
+/// Transform every field of a raw obkv store into a JSON Object.
+pub fn all_obkv_to_json(obkv: &obkv::KvReaderU16, fields_ids_map: &FieldsIdsMap) -> Result<Object> {
+    let all_keys = obkv.iter().map(|(k, _v)| k).collect::<Vec<_>>();
+    obkv_to_json(all_keys.as_slice(), fields_ids_map, obkv)
 }
 
 /// Transform a JSON value into a string that can be indexed.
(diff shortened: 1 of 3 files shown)


milli/src/lib.rs
```diff
@@ -88,7 +88,7 @@ pub fn absolute_from_relative_position(field_id: FieldId, relative: RelativePosi
 pub fn obkv_to_json(
     displayed_fields: &[FieldId],
     fields_ids_map: &FieldsIdsMap,
-    obkv: obkv::KvReaderU16,
+    obkv: &obkv::KvReaderU16,
```
Kerollmops, line 91: As `KvReaderU16` impl `Copy` I am not sure why you changed this function signature. Could you rollback that change as it breaks the API, please?

Kerollmops (changes requested):
Thank you, this looks good to me, it will also help us debug things :)

milli/src/lib.rs
GregoryConrad: Sure thing! That's just a by-product of me still learning Rust--not sure when it's best to rely on clone/copy or just pass by reference. I was figuring passing by reference would help save the copy operation during runtime, but that is probably just a micro-optimisation :).

Kerollmops (approved):
Looks good to me. Thank you very much!
bors merge

Kerollmops:
Hey @GregoryConrad, Would it be possible to quickly fix the Clippy CI in a dedicated commit, please?

Kerollmops:
Hum... strange, it looks like the issue was in the _filter-parser_ crate. Are you sure you use the latest version of Rust? i.e. `rustup update`.

GregoryConrad:
@Kerollmops I still can't replicate:

Screenshot 2022-11-28 at 11 08 31 AM

Would you mind copy-pasting the clippy output you have from `filter-parser` so I can fix it? Perhaps it is on an outdated commit (before I reverted the pass-by-reference changes I made)?

Kerollmops:
Hey @GregoryConrad, sorry for the late reply. I just tried to trigger the Clippy error again and could not output it again 🎉

PR #708, opened 2022-11-24
Reduce memory usage of the MatchingWords structure

loiclec:
# Pull Request

## Related issue
Fixes (partially) https://github.com/meilisearch/meilisearch/issues/3115 

## What does this PR do?
1. Reduces the memory usage caused by the creation of a 10-word query tree by 20x. 
   This is done by deduplicating the `MatchingWord` values, which are heavy because of their inner DFA. The deduplication works by wrapping each `MatchingWord` in a reference-counted box and using a hash map to determine whether a  `MatchingWord` DFA already exists for a certain signature, or whether a new one needs to be built.
 
2. Avoid the worst-case scenario of creating a `MatchingWord` for extremely long words that cannot be indexed by milli.

### DIFF
--- a/milli/src/lib.rs
+++ b/milli/src/lib.rs
@@ -70,6 +70,21 @@ pub type SmallVec8<T> = smallvec::SmallVec<[T; 8]>;
 /// expressed in term of latitude and longitude.
 pub type GeoPoint = rstar::primitives::GeomWithData<[f64; 3], (DocumentId, [f64; 2])>;
 
+/// The maximum length a LMDB key can be.
+///
+/// Note that the actual allowed length is a little bit higher, but
+/// we keep a margin of safety.
+const MAX_LMDB_KEY_LENGTH: usize = 500;
+
+/// The maximum length a field value can be when inserted in an LMDB key.
+///
+/// This number is determined by the keys of the different facet databases
+/// and adding a margin of safety.
+pub const MAX_FACET_VALUE_LENGTH: usize = MAX_LMDB_KEY_LENGTH - 20;
+
+/// The maximum length a word can be
+pub const MAX_WORD_LENGTH: usize = MAX_LMDB_KEY_LENGTH / 2;
+
 pub const MAX_POSITION_PER_ATTRIBUTE: u32 = u16::MAX as u32 + 1;
 
 // Convert an absolute word position into a relative position.
--- a/milli/src/search/matches/matching_words.rs
+++ b/milli/src/search/matches/matching_words.rs
@@ -2,6 +2,7 @@ use std::cmp::{min, Reverse};
 use std::collections::BTreeMap;
 use std::fmt;
 use std::ops::{Index, IndexMut};
+use std::rc::Rc;
 
 use charabia::Token;
 use levenshtein_automata::{Distance, DFA};
@@ -14,11 +15,11 @@ type IsPrefix = bool;
 /// referencing words that match the given query tree.
 #[derive(Default)]
 pub struct MatchingWords {
-    inner: Vec<(Vec<MatchingWord>, Vec<PrimitiveWordId>)>,
+    inner: Vec<(Vec<Rc<MatchingWord>>, Vec<PrimitiveWordId>)>,
 }
 
 impl MatchingWords {
-    pub fn new(mut matching_words: Vec<(Vec<MatchingWord>, Vec<PrimitiveWordId>)>) -> Self {
+    pub fn new(mut matching_words: Vec<(Vec<Rc<MatchingWord>>, Vec<PrimitiveWordId>)>) -> Self {
         // Sort word by len in DESC order prioritizing the longuest matches,
         // in order to highlight the longuest part of the matched word.
         matching_words.sort_unstable_by_key(|(mw, _)| Reverse((mw.len(), mw[0].word.len())));
@@ -35,7 +36,8 @@ impl MatchingWords {
 /// Iterator over terms that match the given token,
 /// This allow to lazily evaluate matches.
 pub struct MatchesIter<'a, 'b> {
-    inner: Box<dyn Iterator<Item = &'a (Vec<MatchingWord>, Vec<PrimitiveWordId>)> + 'a>,
+    #[allow(clippy::type_complexity)]
+    inner: Box<dyn Iterator<Item = &'a (Vec<Rc<MatchingWord>>, Vec<PrimitiveWordId>)> + 'a>,
     token: &'b Token<'b>,
 }
 
@@ -126,7 +128,7 @@ pub enum MatchType<'a> {
 /// Structure helper to match several tokens in a row in order to complete a partial match.
 #[derive(Debug, PartialEq)]
 pub struct PartialMatch<'a> {
-    matching_words: &'a [MatchingWord],
+    matching_words: &'a [Rc<MatchingWord>],
     ids: &'a [PrimitiveWordId],
     char_len: usize,
 }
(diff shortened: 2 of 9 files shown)


milli/src/lib.rs
```diff
@@ -70,6 +70,21 @@ pub type SmallVec8<T> = smallvec::SmallVec<[T; 8]>;
...
+
+/// The maximum length a field value can be when inserted in an LMDB key.
+///
+/// This number is determined by the keys of the different facet databases
+/// and adding a margin of safety.
+pub const MAX_FACET_VALUE_LENGTH: usize = MAX_LMDB_KEY_LENGTH - 20;
+
+/// The maximum length a word can be
+pub const MAX_WORD_LENGTH: usize = MAX_LMDB_KEY_LENGTH / 2;
+
```
loiclec, line 87: This constant has been moved to the root of the crate so that it is accessible within the `search` module, and not only the `update::index_documents` one.

milli/src/search/matches/matching_words.rs
```diff
@@ -91,10 +105,13 @@ impl PartialEq for MatchingWord {
 }
 
 impl MatchingWord {
-    pub fn new(word: String, typo: u8, prefix: IsPrefix) -> Self {
+    pub fn new(word: String, typo: u8, prefix: IsPrefix) -> Option<Self> {
+        if word.len() > MAX_WORD_LENGTH {
+            return None;
+        }
```
loiclec, line 111: Can you think of another condition to add here that would avoid creating more useless matching word DFA?

milli/src/search/query_tree.rs
```diff
@@ -540,6 +544,29 @@ fn create_query_tree(
...
 }
 
+#[derive(Default, Debug)]
+struct MatchingWordCache {
+    all: Vec<Rc<MatchingWord>>,
+    map: HashMap<(String, u8, bool), Rc<MatchingWord>>,
+}
+impl MatchingWordCache {
+    fn insert(&mut self, word: String, typo: u8, prefix: bool) -> Option<Rc<MatchingWord>> {
+        match self.map.entry((word.clone(), typo, prefix)) {
```
loiclec, line 554: This relies on the assumption that two `MatchingWord` structures are perfectly equivalent if they were created with the same `(word: String, typo: u8, prefix: bool)` parameters.

milli/src/search/query_tree.rs
```diff
@@ -562,16 +590,19 @@ fn create_matching_words(
                     for synonym in synonyms {
                         let synonym = synonym
                             .into_iter()
-                            .map(|syn| MatchingWord::new(syn, 0, false))
+                            .flat_map(|syn| matching_word_cache.insert(syn, 0, false))
                             .collect();
```
loiclec, line 594: Now, `matching_word_cache.insert` can return `None` if the length of its string argument exceeds the maximum word length. (If it succeeds instead, it returns an `Option<MatchingWord>`).

Thinking about it more, maybe it would be better to ensure that all the components of the synonyms can have a corresponding `MatchingWord` before creating the list, instead of simply skipping the components that are too long. However, it is probably not that important.

milli/src/search/query_tree.rs
```diff
@@ -580,17 +611,24 @@ fn create_matching_words(
...
                 let ids: Vec<_> =
                     (0..words.len()).into_iter().map(|i| id + i as PrimitiveWordId).collect();
-                let words =
-                    words.into_iter().flatten().map(|w| MatchingWord::new(w, 0, false)).collect();
+                let words = words
+                    .into_iter()
+                    .flatten()
+                    .flat_map(|w| matching_word_cache.insert(w, 0, false))
+                    .collect();
                 matching_words.push((words, ids));
```
loiclec, line 632: Same comment as for `synonym` on line 593.

milli/src/search/query_tree.rs
```diff
@@ -1294,6 +1347,27 @@ mod test {
...
+        insta::assert_snapshot!(format!("{matching_words:?}"), @r###"
+        [
+        ([MatchingWord { word: "house", typo: 1, prefix: true }], [3])
+        ([MatchingWord { word: "house", typo: 1, prefix: true }], [2])
+        ([MatchingWord { word: "whata", typo: 1, prefix: false }], [0, 1])
+        ([MatchingWord { word: "house", typo: 1, prefix: true }], [2])
+        ([MatchingWord { word: "house", typo: 1, prefix: true }], [1])
+        ([MatchingWord { word: "what", typo: 0, prefix: false }], [0])
+        ([MatchingWord { word: "a", typo: 0, prefix: false }], [1])
+        ]
```
loiclec, line 1367: It looks like they are not de-duplicated, but they actually are. Each line describes a `Rc<MatchingWord>`. For example, the first two lines point to the same object in memory.

milli/src/search/query_tree.rs
```diff
@@ -1310,4 +1384,60 @@ mod test {
...
+
+        // Use these snapshots to measure the exact memory usage.
+        // The values below were correct at the time I wrote them.
+        // insta::assert_snapshot!(format!("{}", resident_after - resident_before), @"4486950");
+        // insta::assert_snapshot!(format!("{}", allocated_after - allocated_before), @"7107502");
+
+        // Note, with the matching word cache deactivated, the memory usage was:
+        // insta::assert_snapshot!(format!("{}", resident_after - resident_before), @"91248697");
+        // insta::assert_snapshot!(format!("{}", allocated_after - allocated_before), @"125697588");
+        // or about 20x more resident memory (90MB vs 4.5MB)
```
loiclec, line 1438: This is the where the impact of this PR is measured.

Note that since I had to create a `global_allocator` to measure it, *every other test* of `milli` now also uses this `CountingAlloc`. What are the effects of this? Do we like it, dislike it?

On one hand, it will slow down other tests. On the other hand, we can now measure the memory usage of any operation quite easily.

milli/src/search/query_tree.rs
ManyTheFish: Well, first I'm not sure of what you expect by using `flat_map`:
1) If you want to filter the None values of a synonym and keep all the other values, I would suggest using [`filter_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map).
2) If you want to keep all the values until you encounter a None, I would suggest using [`map_while`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map_while) (it seems equivalent than your current implementation, but I may be wrong).

However, you suggested to keep a synonyms only if all the words are valid, which make sens.
In this case, I think you can collect directly the value in an `Option<Vec<_>>`:
```rust
let synonym: Option<Vec<_>> = synonym
    .into_iter()
    .map(|syn| matching_word_cache.insert(syn, 0, false))
    .collect();
```
(didn't try 😬)

milli/src/search/query_tree.rs
ManyTheFish: so I'll do the same Answer !

milli/src/search/query_tree.rs
```diff
@@ -645,7 +685,7 @@ fn create_matching_words(
                                 for synonym in synonyms {
                                     let synonym = synonym
                                         .into_iter()
-                                        .map(|syn| MatchingWord::new(syn, 0, false))
+                                        .flat_map(|syn| matching_word_cache.insert(syn, 0, false))
```
ManyTheFish, line 688: same as before `flat_map` is obscure 😬

milli/src/search/matches/matching_words.rs
ManyTheFish: I'm afraid no, all I am thinking imply fetching data in the index

ManyTheFish (changes requested):

milli/src/search/query_tree.rs
loiclec: UPDATE: since `cargo_test` runs multiple tests in parallel, this test does not work reliably. I have now commented it out entirely (including the `CountingAlloc`).

milli/src/search/query_tree.rs
ManyTheFish: Are you sure you don't want to use a `map` instead, and collect into an `Option<Vec<_>>` as you did before?
Honestly, I'd prefer to avoid the use of `flat_map`, and at least replace it with a `filter_map` or a `map_while`. 😬

ManyTheFish (changes requested):

ManyTheFish (approved):
Great!

Bors merge

Kerollmops:
Ho no! We should not merge new PRs on this main branch until we release v0.30.1 of Meilisearch and fix the current bugs!
Should we revert this PR on main @curquiza?

PR #710, opened 2022-11-24
Update Clippy to use Rust Stable

Kerollmops:
This PR changes the CI to use Rust stable for Clippy. This way we will reduce the number of times we break the CI. [The version will only change every two months or so](https://www.whatrustisit.com/).

We can't do it for rustfmt yet. See: https://github.com/meilisearch/milli/pull/710#discussion_r1031654931

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -56,7 +56,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: stable
           override: true
           components: clippy
       - name: Cache dependencies
@@ -74,7 +74,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: stable
           override: true
           components: rustfmt
       - name: Cache dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -74,7 +74,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: nightly
+          toolchain: stable
```
irevoire, line 77:
```suggestion
          toolchain: nightly
```

We can't do that for rustfmt since we're using this `.rustfmt.toml` file:
```toml
unstable_features = true

use_small_heuristics = "max"
imports_granularity = "Module"
group_imports = "StdExternalCrate"
```

The CI seems to work, but it actually doesn't; look at the github action; it outputted one thousand times this warning:
```
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
Warning: can't set `unstable_features = true`, unstable features are only available in nightly channel.
```

irevoire (changes requested):

.github/workflows/rust.yml
Kerollmops: Ho you are absolutely right, we should keep nightly for rustfmt. But can't we `-D warnings` for rustfmt too?

Kerollmops: Nope, this option doesn't exists...

irevoire (approved):
bors merge

PR #712, opened 2022-11-30
Fix bulk facet indexing bug

loiclec:
# Pull Request

## Related issue
Fixes (partially, until merged into meilisearch) https://github.com/meilisearch/meilisearch/issues/3165

## What does this PR do?
Fixes a bug where indexing certain numbers of filterable attribute values in bulk led to corrupted facet databases. This was due to a lossy integer conversion which would ultimately prevent entire levels of the facet database to be written into LMDB.

More specifically, this change was made:
```diff
      - if cur_writer_len as u8 >= self.min_level_size {
      + if cur_writer_len >= self.min_level_size as usize {
```
I also checked other comparisons to `min_level_size` and other conversions such as `x as u8` in this part of the codebase.

### DIFF
--- a/milli/src/update/facet/bulk.rs
+++ b/milli/src/update/facet/bulk.rs
@@ -288,6 +288,8 @@ impl<R: std::io::Read + std::io::Seek> FacetsUpdateBulkInner<R> {
                 for bitmap in sub_bitmaps {
                     combined_bitmap |= bitmap;
                 }
+                // The conversion of sub_bitmaps.len() to a u8 will always be correct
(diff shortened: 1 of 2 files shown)


Kerollmops (approved):
Thank you very much for this fix 🎉 Can you confirm that a dump will be required to move to the next version of Meilisearch?

bors merge

loiclec:
> Can you confirm that a dump will be required to move to the next version of Meilisearch?

yes, a dump will be required :(

PR #719, opened 2022-12-03
Add `milli::filter_parser` & `From<&str>` implementation for `Token`

GregoryConrad:
## What does this PR do?
The current `milli::Filter` and `milli::FilterCondition` APIs require working with some members of `filter_parser` directly that `milli::` does *not* re-export to its users. Also, using `filter_parser` does not make sense when using milli from an embedded context where there is no query to parse.

Instead of reworking `milli::Filter` and `milli::FilterCondition`, this PR adds two non-breaking changes that ease the use of milli:
- Re-exports the dependent version of `filter_parser` in `milli`, just like `milli::heed`
- Implements `From<&str>` for `filter_parser::Token`
  - This will also allow some basic tests that need to create a `Token` from a string to avoid some boilerplate.

In conjunction, both of these will allow milli users to easily create a `Token` from a `&str` without needing to add `filter_parser` as an extra dependency.

Note: I wanted to use `FromStr` for the `From` implementation; however, it requires returning a `Result` which is not needed for the conversion. Thus, I just left it as `From<&str>`.

### DIFF
--- a/filter-parser/src/lib.rs
+++ b/filter-parser/src/lib.rs
@@ -115,6 +115,13 @@ impl<'a> From<Span<'a>> for Token<'a> {
     }
 }
 
+/// Allow [Token] to be constructed from &[str]
+impl<'a> From<&'a str> for Token<'a> {
+    fn from(s: &'a str) -> Self {
+        Token::from(Span::new_extra(s, s))
+    }
+}
+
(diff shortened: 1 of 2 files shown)


irevoire (approved):
That's good for me.
I'll let @Kerollmops merge if he's ok with the PR as well.

PR #720, opened 2022-12-05
Make soft deletion optional in document addition and deletion + add lots of tests

loiclec:
# Pull Request

## What does this PR do?
When debugging recent issues, I created a few unit tests in the hopes reproducing the bugs I was looking for. In the end, I didn't find any, but I thought it would still be good to keep those tests. 

More importantly, I added a field to the `DeleteDocuments` and `IndexDocuments` builders, called `disable_soft_deletion`. If set to `true`, the indexing/deletion will never add documents to the `soft_deleted_documents_ids` and instead perform a real deletion of the documents from the databases.

For the new tests, I have:
- Improved the insta-snapshot format of the `external_documents_ids` structure
- Added more tests for the facet DB indexing, deletion, and search algorithms, making sure to test them when the facet DB contains strings (instead of numbers) as well.
- Added more tests for the incremental indexing of the prefix proximity databases. For example, to see if documents are replaced correctly and if common prefixes are deleted correctly.
- Added tests that mix soft deletion and hard deletion, including when processing batches of document updates.


loiclec (commented):
I wrote comments to highlight where the soft deletion changes were made, to focus the reviewing efforts on the consequential parts of the PR.

I also don't expect the reviewer to check the content of the .snap files, of course. You can trust that I have read them and didn't see anything wrong with their content :) Don't forget you can filter the files to review by their extension, and thus exclude all `.snap` files.

milli/src/search/facet/facet_sort_ascending.rs
```diff
@@ -80,6 +80,8 @@ impl<'t, 'e> Iterator for AscendingFacetSort<'t, 'e> {
                 // that we found all the documents in the sub level iterations already,
                 // we can pop this level iterator.
                 if documents_ids.is_empty() {
+                    // break our of the for loop into the end of the 'outer loop, which
```
irevoire, line 83:
```suggestion
                    // break out of the for loop into the end of the 'outer loop, which
```

irevoire (approved):
Nice, thanks for the tests!
bors merge

PR #723, opened 2022-12-06
Fix bug in handling of soft deleted documents when updating settings

loiclec:
# Pull Request

## Related issue
Fixes (partially, until merged into meilisearch) https://github.com/meilisearch/meilisearch/issues/3021

## What does this PR do?
This PR fixes the bug where a `missing key in documents database` internal error message could appear when indexing documents.

When updating the settings, before clearing the database and before creating the transform output, we now modify the `ExternalDocumentsIds` structure to get rid of all references to soft deleted document ids in its FSTs.

It used to be that updating the settings would clear the soft-deleted document ids, but keep the original `ExternalDocumentsIds` structure. As a consequence of this, when processing a future document addition, we could wrongly believe that a document was being replaced when, in fact, it was a completely new document. See the tests `bug_3021_first` and `bug_3021` for a minimal test case that would have reproduced the issue.
 
We need to take special care to:
- evaluate how users should update to v0.30.1 (containing this fix): dump? reimporting all documents from scratch?
- understand IF/HOW this bug could have caused duplicate documents to be returned 
- and evaluate the correctness of the fix of course :)


milli/src/update/index_documents/transform.rs
```diff
@@ -638,17 +646,25 @@ impl<'a, 'i> Transform<'a, 'i> {
...
             original_documents,
             flattened_documents,
-        })
+        };
+
+        let new_facets = output.compute_real_facets(wtxn, self.index)?;
+        self.index.put_faceted_fields(wtxn, &new_facets)?;
+
+        // We clear the full database (words-fst, documents ids and documents content).
+        ClearDocuments::new(wtxn, self.index).execute()?;
```
Kerollmops, line 665: Wonderful!

Kerollmops (approved):
Looks good to me! It is delightful to see this bug finally fixed!
ready to be merged for my part 🧩

milli/src/update/index_documents/transform.rs
```diff
@@ -546,20 +546,28 @@ impl<'a, 'i> Transform<'a, 'i> {
...
-    /// with the attributes reordered accordingly to the `FieldsIdsMap` given as argument.
+    /// Clear all databases. Returns a `TransformOutput` with a file that contains the documents
+    /// of the index with the attributes reordered accordingly to the `FieldsIdsMap` given as argument.
+    ///
     // TODO this can be done in parallel by using the rayon `ThreadPool`.
-    pub fn remap_index_documents(
+    pub fn prepare_for_documents_reindexing(
         self,
-        wtxn: &mut heed::RwTxn,
+        wtxn: &mut heed::RwTxn<'i, '_>,
```
irevoire, line 555:
```suggestion
        wtxn: &mut heed::RwTxn,
```

I don't think we need these lifetimes?

irevoire (approved):
I don't know if we can remove these lifetimes.
But if we can't, you can merge!

milli/src/update/index_documents/transform.rs
Kerollmops: I think we do. I struggled to remove them. Rust is too dumb to elide them...

loiclec: yea, smart enough to suggest them, not enough to elide them

PR #728, opened 2022-12-07
Add some integration tests on the sort criterion

loiclec:
This is simply an integration test ensuring that the sort criterion works properly. 

However, only one version of the algorithm is tested here (the iterative one). To test the version that uses the facet DB, one has to manually set the `CANDIDATES_THRESHOLD` constant to `0`. I have done that and ensured that the test still succeeds. However, in the future, we will probably want to have an option to force which algorithm is used at runtime, for testing purposes.


ManyTheFish (approved):
Thank you for the PR,
bors merge

PR #732, opened 2022-12-08
Interpret synonyms as phrases

loiclec:
# Pull Request

## Related issue
Fixes (when merged into meilisearch) https://github.com/meilisearch/meilisearch/issues/3125

## What does this PR do?
We now map multi-word synonyms to phrases instead of loose words. Such that the request:
```
btw I am going to nyc soon
```
is interpreted as (when the synonym interpretation is chosen for both `btw` and `nyc`):
```
"by the way" I am going to "New York City" soon
```
instead of:
```
by the way I am going to New York City soon
```

This prevents queries containing multi-word synonyms to exceed to word length limit and degrade the search performance.

In terms of relevancy, there is a debate to have. I personally think this could be considered an improvement, since it would be strange for a user to search for:
```
good DIY project
```
and have a result such as:
```
{
    "text": "whether it is a good project to do, you'll have to decide for yourself"
}
```
However, for synonyms such as `NYC -> New York City`, then we will stop matching documents where `New York` is separated from `City`. This is however solvable by adding an additional mapping: `NYC -> New York`.

## Performance

With the old behaviour, some long search requests making heavy uses of synonyms could take minutes to be executed. This is no longer the case, these search requests now take an average amount of time to be resolved.

### DIFF
--- a/milli/src/search/query_tree.rs
+++ b/milli/src/search/query_tree.rs
@@ -339,18 +339,18 @@ fn typos(word: String, authorize_typos: bool, config: TypoConfig) -> QueryKind {
 /// and create the list of operations for the query tree
 fn synonyms(ctx: &impl Context, word: &[&str]) -> heed::Result<Option<Vec<Operation>>> {
     let synonyms = ctx.synonyms(word)?;
-
     Ok(synonyms.map(|synonyms| {
         synonyms
             .into_iter()
             .map(|synonym| {
-                let words = synonym
-                    .into_iter()
-                    .map(|word| {
-                        Operation::Query(Query { prefix: false, kind: QueryKind::exact(word) })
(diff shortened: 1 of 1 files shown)


Kerollmops (approved):
Easy peasy, lemon squeezy 🍋
Let's merge that. Nice job again!

curquiza:
Checked with @gmourier, since this PR has a product impact and we don't get any approval from the product team yet on this, don't merge it until @gmourier validates it 😇

Kerollmops:
Awaiting the approval of @gmourier ⌛

gmourier:
I'm late to the party to signal the product team stamp of approval🏃‍♂️ Thanks for that enhancement @loiclec!

loiclec:
bors merge

PR #735, opened 2022-12-12
Write a limited fuzz test stressing combinations of index operations

loiclec:
# Pull Request

## What does this PR do?
Add a fuzz test that applies multiple operations (document import, document deletion, setting update, clear documents) to an index and verify that none of them crash.


loiclec:
Note: this PR will stay in a draft state for a long time probably, but I'd like to keep it alive as a reminder that fuzz-testing indexing is possible and not that complicated. We can build on it in the future.

PR #737, opened 2022-12-12
Fix typo initial candidates computation

ManyTheFish:
When `Typo` criterion was after a different criterion than `Words` and the previous criterion wasn't returning any candidates at the first iteration of the bucket sort, then the `initial_candidates` were lost.

Now, `Typo`ensure to keep the `initial_candidates` between iterations.

related to https://github.com/meilisearch/meilisearch/issues/3200#issuecomment-1345179578
related to https://github.com/meilisearch/meilisearch/issues/3228

### DIFF
--- a/milli/src/search/criteria/typo.rs
+++ b/milli/src/search/criteria/typo.rs
@@ -143,8 +143,8 @@ impl<'t> Criterion for Typo<'t> {
                     }) => {
                         self.initial_candidates =
                             match (self.initial_candidates.take(), initial_candidates) {
-                                (Some(self_bc), Some(parent_bc)) => Some(self_bc | parent_bc),
-                                (self_bc, parent_bc) => self_bc.or(parent_bc),
+                                (Some(self_ic), Some(parent_ic)) => Some(self_ic | parent_ic),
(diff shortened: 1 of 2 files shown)


irevoire (approved):
Nice, from what I understand it works.

I looked at all the other criterion and it look like some (proximity, geo, arc_desc, attribute, exactness) are doing this;
```rust

                        match initial_candidates {
                            Some(initial_candidates) => {
                                self.initial_candidates |= initial_candidates
                            }
                            None => self.initial_candidates.map_inplace(|c| c | &candidates),
                        }
```

While words is doing what you wrote here.
Is there a reason to chose one method over the other?

Kerollmops (approved):
Looks amazing to me!
bors merge

ManyTheFish:
@irevoire, this is because `Words` and `Typo` criteria contain an `Option<InitialCandidates>` instead of `InitialCandidates`, so I have to make a different `match`.
This comes from the fact that these two criteria are "lazier" than the others because they may not compute any candidates at their step letting the following criterion do it for them.

ManyTheFish:
Poke @curquiza about this fix.

curquiza:
Is it normal we haven't added any tests here?

PR #750, opened 2022-12-19
Fix hard-deletion of an external id that was soft-deleted and then reimported - main

loiclec:
# Pull Request

## Related issue
Fixes (when merged into meilisearch) https://github.com/meilisearch/meilisearch/issues/3021

## What does this PR do?
There was a bug happening when:

1. Documents were added
2. Some of these documents were replaced using soft-deletion
3. A deletion of another non-replaced document takes place and triggers a hard-deletion
4. Documents with the same identifiers as the replaced documents are added again

Then, search results would return duplicate documents. No crash would happen at any time (this is the reason it wasn't caught by the previous fuzz test. I have updated the new one such that it also checks the result of a placeholder search request, which then finds the bug immediately).

The cause of the bug is: 

1. When a hard-deletion is triggered, we try to retrieve the external document id associated with each soft-deleted document id. 
2. Then, we take this list of external document ids and remove each of them from the `ExternalDocumentsIds` structure. 
3. However, this is not correct in case an existing (non-deleted) document shares the external id of a soft-deleted document. 
   
## Implementation of the fix
1. Before we process a permanent deletion, we update the list of soft-deleted document ids.
2. Then, the permanent deletion's job is to remove the soft-deleted documents from all data structures. Therefore, to update `ExternalDocumentsIds`, we can simply call the `delete_soft_deleted_documents_ids_from_fsts` method, which is faster and simpler.

## Correctness
A unit test was added to reproduce the bug. The new fuzz test, when adjusted to check the correctness of a placeholder search, could also instantly reproduce the bug, but now does not find any other problem.

### DIFF
--- a/milli/src/external_documents_ids.rs
+++ b/milli/src/external_documents_ids.rs
@@ -47,30 +47,6 @@ impl<'a> ExternalDocumentsIds<'a> {
         }
     }
 
-    pub fn delete_ids<A: AsRef<[u8]>>(&mut self, other: fst::Set<A>) -> fst::Result<()> {
-        let other = fst::Map::from(other.into_fst());
-        let union_op = self.soft.op().add(&other).r#union();
-
-        let mut iter = union_op.into_stream();
-        let mut new_soft_builder = fst::MapBuilder::memory();
-        while let Some((external_id, docids)) = iter.next() {
-            if docids.iter().any(|v| v.index == 1) {
-                // If the `other` set returns a value here it means
-                // that it must be marked as deleted.
-                new_soft_builder.insert(external_id, DELETED_ID)?;
-            } else {
-                let value = docids.iter().find(|v| v.index == 0).unwrap().value;
-                new_soft_builder.insert(external_id, value)?;
-            }
-        }
-
-        drop(iter);
-
-        // We save this new map as the new soft map.
-        self.soft = new_soft_builder.into_map().map_data(Cow::Owned)?;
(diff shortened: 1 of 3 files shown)


milli/src/external_documents_ids.rs
```diff
@@ -47,28 +48,40 @@ impl<'a> ExternalDocumentsIds<'a> {
...
-                let value = docids.iter().find(|v| v.index == 0).unwrap().value;
-                new_soft_builder.insert(external_id, value)?;
+            // prefer selecting the ids from soft, always
+            let id = indexed_last_value(docids).unwrap();
+
+            if id != DELETED_ID
+                && !self.soft_deleted_docids.contains(id as u32)
+                && other.get(external_id).map_or(true, |b| !b.contains(id as u32))
+            {
+                new_hard_builder.insert(external_id, id)?;
```
irevoire, line 74: Thanks, this is way easier to follow, and if I follow correctly, should be faster :tada:

irevoire (dismissed):
I think we're good to go :+1: 

But I believe you'll need to rewrite some of your tests since I merged this PR: https://github.com/meilisearch/milli/pull/747

That update the way the `index.index_documents_config` works sorry :grimacing:

irevoire (approved):
Thanks a lot; that's way better than the first implementation we thought of initially!

bors merge

loiclec:
Yes, the fuzz test was done on this branch initially :-)

PR #752, opened 2022-12-19
Simplify primary key inference

dureuill:
# Pull Request

## Related issue
Related to https://github.com/meilisearch/meilisearch/issues/3233

### User PoV

- Change primary key inference to only consider a value as a candidate when it ends with "id", rather than when it simply contains "id".
- Change primary key inference to always fail when there are multiple candidates.
- Replace UserError::MissingPrimaryKey with `UserError::NoPrimaryKeyCandidateFound` and `UserError::MultiplePrimaryKeyCandidatesFound`

### Implementation-wise

- Remove uses of UserError::MissingPrimaryKey not pertaining to inference. This introduces a possible panicking path.

## PR checklist

### DIFF
--- a/milli/src/error.rs
+++ b/milli/src/error.rs
@@ -130,8 +130,10 @@ only composed of alphanumeric characters (a-z A-Z 0-9), hyphens (-) and undersco
     MissingDocumentId { primary_key: String, document: Object },
     #[error("Document have too many matching `{}` attribute: `{}`.", .primary_key, serde_json::to_string(.document).unwrap())]
     TooManyDocumentIds { primary_key: String, document: Object },
-    #[error("The primary key inference process failed because the engine did not find any fields containing `id` substring in their name. If your document identifier does not contain any `id` substring, you can set the primary key of the index.")]
-    MissingPrimaryKey,
+    #[error("The primary key inference process failed because the engine did not find any field ending with `id` in its name. Please specify the primary key manually using the `primaryKey` query parameter.")]
+    NoPrimaryKeyCandidateFound,
+    #[error("The primary key inference process failed because the engine found {} fields ending with `id` in their name, such as '{}' and '{}'. Please specify the primary key manually using the `primaryKey` query parameter.", .candidates.len(), .candidates.get(0).unwrap(), .candidates.get(1).unwrap())]
+    MultiplePrimaryKeyCandidatesFound { candidates: Vec<String> },
     #[error("There is no more space left on the device. Consider increasing the size of the disk/partition.")]
     NoSpaceLeftOnDevice,
     #[error("Index already has a primary key: `{0}`.")]
--- a/milli/src/update/index_documents/enrich.rs
+++ b/milli/src/update/index_documents/enrich.rs
@@ -58,17 +58,36 @@ pub fn enrich_documents_batch<R: Read + Seek>(
             }
         },
         None => {
-            let guessed = documents_batch_index
+            let mut guesses: Vec<(u16, &str)> = documents_batch_index
                 .iter()
-                .filter(|(_, name)| name.to_lowercase().contains(DEFAULT_PRIMARY_KEY))
-                .min_by_key(|(fid, _)| *fid);
-            match guessed {
-                Some((id, name)) => PrimaryKey::flat(name.as_str(), *id),
-                None if autogenerate_docids => PrimaryKey::flat(
+                .filter(|(_, name)| name.to_lowercase().ends_with(DEFAULT_PRIMARY_KEY))
+                .map(|(field_id, name)| (*field_id, name.as_str()))
+                .collect();
+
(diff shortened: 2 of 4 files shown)


milli/src/update/index_documents/enrich.rs
```diff
@@ -49,22 +53,34 @@ pub fn enrich_documents_batch<R: Read + Seek>(
...
-                .filter(|(_, name)| name.to_lowercase().contains(DEFAULT_PRIMARY_KEY))
-                .min_by_key(|(fid, _)| *fid);
-            match guessed {
-                Some((id, name)) => PrimaryKey::flat(name.as_str(), *id),
-                None if autogenerate_docids => PrimaryKey::flat(
+                .filter(|(_, name)| name.to_lowercase().ends_with(DEFAULT_PRIMARY_KEY))
+                .map(|(field_id, name)| (*field_id, name.as_str()))
+                .collect();
+
+            guesses.sort_by_key(|(field_id, _)| *field_id);
```
irevoire, line 67: Do you think it's better to sort by `field_id` or by string directly?

I feel like most of the time the most important key is the smallest one (like `id`) instead of the first appearing in the document what do you think?

milli/src/error.rs
```diff
@@ -130,8 +130,10 @@ only composed of alphanumeric characters (a-z A-Z 0-9), hyphens (-) and undersco
     MissingDocumentId { primary_key: String, document: Object },
     #[error("Document have too many matching `{}` attribute: `{}`.", .primary_key, serde_json::to_string(.document).unwrap())]
     TooManyDocumentIds { primary_key: String, document: Object },
-    #[error("The primary key inference process failed because the engine did not find any fields containing `id` substring in their name. If your document identifier does not contain any `id` substring, you can set the primary key of the index.")]
-    MissingPrimaryKey,
+    #[error("The primary key inference process failed because the engine did not find any field ending with `id` in its name. Please specify the primary key manually using the `primaryKey` query parameter.")]
+    NoPrimaryKeyCandidateFound,
+    #[error("The primary key inference process failed because the engine found {} fields ending with `id` in their name. Candidates include '{}', '{}', and more. Please specify the primary key manually using the `primaryKey` query parameter.", .candidates.len(), .candidates.get(0).unwrap(), .candidates.get(1).unwrap())]
+    MultiplePrimaryKeyCandidatesFound { candidates: Vec<String> },
```
irevoire, line 136: While integrating milli in Meili, we should add new tests on these error messages.

irevoire (approved):
Very nice; I left a little comment, but if you think it's ok like that, we can merge :+1:

milli/src/update/index_documents/enrich.rs
dureuill: I don't have a strong opinion one way or another. I chose the order of appearance so that users can find the options easily in the JSON even if the document is huge since the options will tend to be by the top, but even this is a heuristic and I don't mind changing for the lexicographic sort.

dureuill: The logic might be a bit more complicated though because we will have to sort first by length, then lexicographically, to ensure a consistent order.

irevoire: We could sort lexicographically by the end :joy:  
```rust
guesses.sort_by(|left, right| left.chars().rev().partial_cmp(right.chars().rev()).unwrap_or(Ordering::Equal));
```
https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.partial_cmp

Honestly, it's as you wish, it'll be slower for sure, but I don't think that's a big deal personally since we were ready to run an entire indexing process.

dureuill: Changed the sort to:

```rust
            // sort the keys in a deterministic, obvious way, so that fields are always in the same order.
            guesses.sort_by(|(_, left_name), (_, right_name)| {
                // shortest name first
                left_name.len().cmp(&right_name.len()).then_with(
                    // then alphabetical order
                    || left_name.cmp(right_name),
                )
            });
```

Tell me what you think

dureuill:
I added a log looking like this:

```
[2022-12-21T09:19:42Z INFO  milli::update::index_documents::enrich] Primary key was not specified in index. Inferred to 'id'
```

when inference took place

milli/src/error.rs
dureuill: See https://github.com/meilisearch/meilisearch/pull/3269/files#diff-8c6bf09fe3009f0d2a301764e29f7ec3bffe0d5fbca17af2072b13df1ac4786aR900

milli/src/update/index_documents/enrich.rs
irevoire: yep, imo, that’s perfect 👌

irevoire (approved):
bors merge

PR #761, opened 2023-01-11
Integrate deserr

loiclec:
1. `Setting<T>` now implements `DeserializeFromValue`
2. The settings now store ranking rules as strongly typed `Criterion` instead of `String`, since the validation of the ranking rules will be done on meilisearch's side from now on


irevoire (approved):
noice thank you
bors merge

PR #764, opened 2023-01-17
Update deserr to latest version

loiclec:
Update deserr to 0.1.5, which changes the `DeserializeFromValue` trait, getting rid of the `default()` method.


irevoire (approved):
bors merge

PR #772, opened 2023-01-24
Throw an error on unknown fields specified in the _geo field

irevoire:
Fix parts of https://github.com/meilisearch/meilisearch/issues/3414

### DIFF
--- a/milli/src/index.rs
+++ b/milli/src/index.rs
@@ -2326,4 +2326,35 @@ pub(crate) mod tests {
 
         db_snap!(index, geo_faceted_documents_ids); // ensure that no more document was inserted
     }
+
+    #[test]
+    fn unexpected_extra_fields_in_geo_field() {
+        let index = TempIndex::new();
+
+        index
+            .update_settings(|settings| {
+                settings.set_primary_key("id".to_string());
+                settings.set_filterable_fields(HashSet::from(["_geo".to_string()]));
+            })
+            .unwrap();
+
+        let err = index
+            .add_documents(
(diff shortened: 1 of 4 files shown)


milli/src/index.rs
```diff
@@ -2326,4 +2326,35 @@ pub(crate) mod tests {
...
+                settings.set_filterable_fields(HashSet::from(["_geo".to_string()]));
+            })
+            .unwrap();
+
+        let err = index
+            .add_documents(
+                documents!({ "id" : "doggo", "_geo": { "lat": 1, "lng": 2, "doggo": "are the best" }}),
+            )
+            .unwrap_err();
+        insta::assert_display_snapshot!(err, @r###"The `_geo` field in the document with the id: `"\"doggo\""` contains the following unexpected fields: `{"doggo":"are the best"}`."###);
```
loiclec, line 2346:
```rust
`{"doggo":"are the best"}`
```
This could be a bit confusing, because this object appears nowhere in the original document. I don't know what the alternative should be though, and it's just an error message, so not a big deal.

loiclec (approved):
Looks good, thank youuuu :)

milli/src/index.rs
irevoire: Yeah I thought of displaying only a list of field name, but then I found it a bit confusing as well when it comes to more complex values 😩 

I guess we’ll wait and see, if people complains about this message we’ll change it.

irevoire:
bors merge

PR #779, opened 2023-02-22
Fix phrase search containing only stop words

ManyTheFish:
# Summary

A search with a phrase containing only stop words was returning an HTTP error 500,
this PR filters the phrase containing only stop words dropping them before the search starts, a query with a phrase containing only stop words now behaves like a placeholder search.

fixes https://github.com/meilisearch/meilisearch/issues/3521 for v1.0.2

related v1.1.0 PR on Meilisearch: https://github.com/meilisearch/meilisearch/pull/3525


irevoire (approved):
Nice thank you 🛼 
bors merge

